PR #2, opened 2015-03-24
Make config parser keys case sensitive

leonhandreke:
Took me while to figure out it wasn't I think it's generally less confusing this way.


dbkr:
Sure, lgtm.

PR #4, opened 2016-10-18
Kill off gcm-client

dbkr:
It doesn't support setting the 'priority' field and offers little to no value over just doing the HTTP pokes ourselves, so... do it ourselves.


NegativeMjark:
LGTM

PR #5, opened 2016-11-09
Change gevent.wait to gevent.sleep

dbkr:
wait without objects means wait for the event loop to finish,
which only makes sense from the main greenlet. We want
gevent.sleep().


erikjohnston:
LGTM

PR #6, opened 2016-11-09
'MessageTooBig' is a permanent failure

dbkr:


erikjohnston:
LGTM

erikjohnston:
Although, it may be nice to resend with a smaller content? Or something? Android doesn't really care about the content iirc?

dbkr:
Yeah, although we know the limit is 2kb, so we may as well do what we do in iOS and trim it down to 2kb in advance. Also I think android does still use the content sometimes.

PR #7, opened 2016-11-09
Don't flatten the content dict in GCM

dbkr:
Since the flattening code neglected to delete the original nested
keys, we were sending the body and msgtype twice, so we were
often exceeding the message size limit unnecessarily.

By the looks of it, GCM has in fact always supported nested keys,
contrary to the comment (although the google library on Android
exposes the data dict as Map<String, String>). Vector android
uses content.body (ie. the nested key), so the flattened keys
have never been used (console uses neither).


erikjohnston:
lgtm

PR #8, opened 2016-11-09
Don't reject pushkeys because message is too long

dbkr:
Permanent failure for reg id != permanent failure for message


erikjohnston:
LGTM

Issue #14, opened 2017-01-30
MQTT Push Provider?


jcgruenhage:
MQTT would be an alternative for Riot on Android/maybe iOS, Facebook is using it for their messenger instead of GCM/APNS. This way Google and Apple would get less metadata, which is one of the reasons people prefer self hosted open source solutions.

If this is not in the scope for sygnal, maybe someone could implement another push gateway that would do this.

ara4n:
That would be awesome!

jcgruenhage:
I fear I just already have too many projects I am working on at the moment, so someone else would need to do that :D

awesome-manuel:
Would be nice if LineageOS and others could establish their own standard for push notifications. Should be based on web push (https://developers.google.com/web/fundamentals/push-notifications/) IMHO.

jcgruenhage:
@awesome-manuel you might want to look at https://bubu1.eu/openpush/

minecraftchest1:
Unified Push (unifiedpush.org) is a nother interesting project. It already supports matrix using their common gatwyays project which implements the push notification api.

awesome-manuel:
@jcgruenhage @minecraftchest1 you're right, but that means something like sygnal (or any other service that want's to support push) has to implement Firebase, Apple APNs, Web Push, Openpush and Unified Push. Actually Web Push should be enough and all the others could be implemented on top of it.

matrixbot:
For your information, this issue has been copied over to the Element fork of `sygnal`: <https://github.com/element-hq/sygnal/issues/14>

PR #18, opened 2017-06-13
[WIP] MQTT support

jooert:
This is a first stab at MQTT support (see #14). I tried it with a local mosquitto as the broker and it basically just works, although it's probably not very robust. Open questions for me are:

- What data exactly should be sent over MQTT? Currently, the code uses the same `build_data` method as `GcmPushkin` did to build a dictionary, dumps it as JSON and publishes it over MQTT.
- Under which topic should a notification be published? Currently, we just publish it to a topic named the same as the pushkey for the device. I could then imagine devices that want to receive the notification to connect to the broker with the pushkey as their client id while the broker should be configured such that clients can only subscribe to topics that match their client id. That way only who knows the pushkey for a specific device can subscribe to the notifications. (I'm not sure if the pushkey is meant to be used like that...)
- Which options should be exposed in `sygnal.conf`? An MQTT broker could run on a different port and most brokers also allow authenticating using a client certificate instead of username and password.

This is my first contact with MQTT so I could be entirely off-track. Thus, I would love to hear from people that have more experience with it. :wink:

@jcgruenhage, @ara4n

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ setup(
         "gevent>=1.0.1",
         "pushbaby>=0.0.9",
         "grequests",
+        "paho-mqtt",
     ],
     long_description=read("README.rst"),
 )
--- a/sygnal.conf.sample
+++ b/sygnal.conf.sample
@@ -7,3 +7,7 @@ com.example.myapp.ios.type = apns
 com.example.myapp.ios.certfile = com.example.myApp_prod_APNS.pem
 com.example.myapp.android.type = gcm
 com.example.myapp.android.apiKey = your_api_key_for_gcm
+com.example.myapp.mqtt.type = mqtt
+com.example.myapp.mqtt.broker = mqtt_broker
+com.example.myapp.mqtt.username = mqtt_username
+com.example.myapp.mqtt.password = mqtt_password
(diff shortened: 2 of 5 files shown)


ara4n:
oooooooh! this looks very very cool!  sygnal is @dbkr's baby, so i suspect he has some thoughts here.  The data sent to GCM or APNS is a very good start - the alternative would be to poke the app to wake up and do a sync itself, but if we can avoid doing that it'll certainly save battery & bandwidth.

Have you tried hooking it into a client (e.g. matrix-android-sdk) yet?

jooert:
> The data sent to GCM or APNS is a very good start - the alternative would be to poke the app to wake up and do a sync itself, but if we can avoid doing that it'll certainly save battery & bandwidth.

MQTT allows arbitrary data to be send, so we can just send the notification instead of only poking the client. However, we probably have to fix the format of the notification data to be sent. (That's probably a reason why it would be good to implement client support in matrix-android-sdk instead of Riot itself.)

> Have you tried hooking it into a client (e.g. matrix-android-sdk) yet?

No, not yet, although that's definitely something I want to try out.

dbkr:
This is cool! I have no experience with MQTT personally but it looks plausible from a sygnal point-of-view. Is there an example of a client that uses it?

jooert:
No, although I would love to see Riot on Android supporting it (relevant issue is https://github.com/vector-im/riot-android/issues/1066).

richvdh:
this seems to be bitrotting

PR #20, opened 2017-08-22
ApnsPushkin improvements

morozkin:
Signed-off-by: Denis Morozov dmorozkn@gmail.com

### DIFF
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -168,7 +168,19 @@ class ApnsPushkin(Pushkin):
                     loc_args = [from_display]
 
         elif n.type == 'm.call.invite':
-            loc_key = 'VOICE_CALL_FROM_USER'
+            is_video_call = False
+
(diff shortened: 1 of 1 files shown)


dbkr:
This changes pushes for image messages from, "Bob in Cats: image.jpg" to "Bob posted a picture in Cats"

@ara4n @lampholder @AmandineLP yay or nay? Personally I like having that little bit more information about what the image might be (eg. coffee.gif)

ara4n:
yeah, i would prefer to keep the filename or other details too...

morozkin:
@dbkr wdyt?

dbkr:
Yep, let's do that then. The annoyance is that if the string isn't in the strings file, it'll just display the raw string sent by the server, so I think we'll have to make sure everyone has updated before we start sending the new string.

giomfo:
FYI the strings have been updated on client side (on develop)
https://github.com/vector-im/riot-ios/pull/1486

PR #21, opened 2017-09-07
Fix GCM push response access to enable logging again

krombel:
The response from GCM/FCM was referred to wrongly so the operation failed (and could not log) the response. This caused an Exception and therefore a 500 response from sygnal. And this let synapse to try pushing again when (hopefully) the config of sygnal got fixed up again.

This PR aims to enable logging but also return 500 when sygnal got configured wrongly.


dbkr:
Looks sensible - can you sign off as per https://github.com/matrix-org/sygnal/blob/master/CONTRIBUTING.rst please?

PR #23, opened 2018-04-17
Don't try to get retry-after if no response

dbkr:
We'd get here if the connection failec/ timed out, so check there
is actually a response before looking for a retry-after header.


dbkr:
yep, it is

PR #25, opened 2018-04-17
Limit field length in APNS

dbkr:
Pushbaby will truncate the fields until it fits in 2KB, but in a
fairly slow and unoptimised way, so if it's way over, it can take
forever.


sygnal/apnspushkin.py
```diff
@@ -38,6 +38,7 @@
 CREATE UNIQUE INDEX IF NOT EXISTS b64token on apns_failed(b64token);
 """
 
+MAX_FIELD_LENGTH = 1024
```
richvdh: A comment wouldn't hurt. more-or-less what you put in the commit comment.

PR #26, opened 2018-06-14
initial commit with dummy_receiver

BACtaki:
TODO

*fix pushkey pushdown to rabbitmq receiver
*replace GCM with custom async receiver
*dummy_receiver to android


richvdh:
this seems to be bitrotting

PR #32, opened 2019-03-02
Basic python3 support.

### LINKED ISSUE #28: Python 3 Upgrade
Now that synapse supports python 3, what about sygnal?

Looking at the dependencies, flask supports python 3 on the currently pinned version. However  gevent (now pinned at 1.0.1) will need to be at least 1.1. pushbaby also needs an update, but this shouldn't be difficult as it only depends on gevent.

evilham:
This has been tested to work fine with FB / GCM, apns may need further
changes.

Fixes #28.

### DIFF
--- a/sygnal/__init__.py
+++ b/sygnal/__init__.py
@@ -15,10 +15,8 @@
 # limitations under the License.
 
 
-try:
-    import ConfigParser as configparser
-except ImportError:
-    import configparser
+from six.moves import configparser
+
 import json
 import logging
 import sys
@@ -174,10 +172,7 @@ class ClientError(Exception):
 
 
 def parse_config():
-    if hasattr(configparser, 'SafeConfigParser'):
-        cfg = configparser.SafeConfigParser(CONFIG_DEFAULTS)
-    else:
-        cfg = configparser.ConfigParser(CONFIG_DEFAULTS)
+    cfg = configparser.SafeConfigParser(CONFIG_DEFAULTS)
     # Make keys case-sensitive
     cfg.optionxform = str
     for sect in CONFIG_SECTIONS:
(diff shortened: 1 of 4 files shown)


sygnal/__init__.py
```diff
@@ -15,7 +15,10 @@
 # limitations under the License.
 
 
-import ConfigParser
+try:
+    import ConfigParser as configparser
+except ImportError:
+    import configparser
```
erikjohnston: You can use `six` to do this: `from six.moves import configparser`

sygnal/__init__.py
```diff
@@ -171,13 +174,16 @@ class ClientError(Exception):
 
 
 def parse_config():
-    cfg = ConfigParser.SafeConfigParser(CONFIG_DEFAULTS)
+    if hasattr(configparser, 'SafeConfigParser'):
+        cfg = configparser.SafeConfigParser(CONFIG_DEFAULTS)
+    else:
+        cfg = configparser.ConfigParser(CONFIG_DEFAULTS)
```
erikjohnston: The `six` configparse always has `SafeConfigParser

sygnal/db.py
```diff
@@ -17,15 +17,18 @@
 import sqlite3
 import logging
 import threading
-import Queue
+try:
+    import Queue as queue
+except ImportError:
+    import queue
```
erikjohnston: `from six.moves import queue`

evilham:
Wonderful, I wasn't sure if pulling six was a desired thing, and since there wasn't much that needed touching...

Will take a go at it soon.

erikjohnston:
Ah, cool, yeah I don't think pulling in `six` is a big deal tbh.

evilham:
This would be ready for another review.

erikjohnston:
Awesome! Thank you

PR #34, opened 2019-04-25
Bump flask version

richvdh:
flask 0.10.0 is pretty old and has known security vulns.

I'm not sure how to test this; it starts and responds to GET requests...


poljar:
Testing flask i pretty straightforward. You just create a test client from the app with `app.test_client()` and can call the `get()` method on that client.

More info [here](http://flask.pocoo.org/docs/1.0/testing/#the-testing-skeleton).

richvdh:
yeahh, but I'm not sure I'm going to go to the effort of creating a test rig right now.

PR #37, opened 2019-05-22
Add basic sentry/prometheus metrics

erikjohnston:


sygnal/__init__.py
```diff
@@ -287,6 +314,17 @@ def setup():
...
+    if cfg.has_option("metrics", "sentry_dsn"):
+        # Only import sentry if enabled
+        import sentry_sdk
+        sentry_sdk.init(
+            dsn=cfg.get("metrics", "sentry_dsn"),
+            integrations=[sentry_sdk.integrations.flask.FlaskIntegration()],
+        )
+
+    if cfg.has_option("metrics", "prometheus_port"):
+        prometheus_client.start_http_server(cfg.get_int("metrics", "prometheus_port"))
```
michaelkaye: SHould include address to listen on pls :/

sygnal/__init__.py
```diff
@@ -26,16 +26,36 @@
...
+
 import sygnal.db
 from sygnal.exceptions import InvalidNotificationException
 
+
+NOTIFS_RECEIVED_COUNTER = Counter(
+    "sygnal_notifications_received", "Number of notification pokes received",
+)
+
+NOTIFS_RECEIVED_COUNTER = Counter(
```
richvdh: one of these needs a different name

sygnal/__init__.py
```diff
@@ -26,16 +26,36 @@
...
+    "sygnal_notifications_devices_received",
+    "Number of devices been asked to push",
+)
+
+NOTIFS_BY_PUSHKIN = Counter(
+    "sygnal_per_pushkin_type",
+    "Number of pushes sent via type of pushkin",
```
richvdh:
```suggestion
    "Number of pushes sent via each type of pushkin",
```

sygnal/__init__.py
```diff
@@ -287,6 +314,17 @@ def setup():
     else:
         logging.basicConfig()
 
+    if cfg.has_option("metrics", "sentry_dsn"):
```
richvdh: is there not some "default config" or "sample config" that this needs adding to?

erikjohnston: I've added it to sample config.

PR #39, opened 2019-05-27
Read config file location from environment vars

### LINKED ISSUE #29: Parameterize location of the config file
Being able to specify the location of the config file would be really great. Currently configuring sygnal is less than ideal as you have to worry about the working directory.

The [current parsing code](https://github.com/matrix-org/sygnal/blob/master/sygnal/__init__.py#L156-L160) has a comment stating this hasn't been implemented because it is awkward when using with gunicorn. 

## Method 1: gunicorn passing
The "correct" way to do this with gunicorn is like so:

```
gunicorn -c /etc/sygnal/gunicorn_config.py sygnal:setup("path/to/config")
```

Then just change the function [`setup`](https://github.com/matrix-org/sygnal/blob/master/sygnal/__init__.py#L223) to take the config path as a param, and plumb it down to `parse_config()`

This is ugly, I agree, but it is the right way according to gunicorn.

## Method 2: environment variable
Alternatively, and perhaps more elegantly, the config could be passed via environment variable. So [the line at](https://github.com/matrix-org/sygnal/blob/master/sygnal/__init__.py#L159)

```python
cfg.read("sygnal.conf")
```

Could become:

```python
cfg.read(os.getenv('SYGNAL_CONFIG') or "sygnal.conf")
```

abeluck:
* Allows conf file path to be specified via the `SYGNAL_CONF` env var
* fixed the README which was using an incorrect name for the conf file

This will have no affect on existing deployments.

Fixes #29

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,10 +18,10 @@ sygnal can use gunicorn's extra hook to perform a clean shutdown which tries as
 hard as possible to ensure no messages are lost.
 
 There are two config files:
- * sygnal.cfg (The app-specific config file)
+ * sygnal.conf (The app-specific config file)
  * gunicorn_config.py (gunicorn's config file)
 
-sygnal.cfg contains configuration for sygnal itself. This includes the location
+sygnal.conf contains configuration for sygnal itself. This includes the location
 and level of sygnal's log file. The [apps] section is where you set up different
 apps that are to be handled. Keys in this section take the form of the app_id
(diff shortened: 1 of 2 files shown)


erikjohnston:
Sure, I guess this works!

Can you read https://github.com/matrix-org/sygnal/blob/master/CONTRIBUTING.rst and sign off your PR please? Then I'll merge

abeluck:
Done.

Thanks @erikjohnston

PR #40, opened 2019-06-12
Add prometheus_client as dependency

### LINKED ISSUE #38: prometheus_client not in dependency list
`prometheus_client` should be added  to `install_requires` in setup.py, or conditionally imported.

abeluck:
ref: https://pypi.org/project/prometheus_client/
fixes #38


erikjohnston:
Woops! Thanks

PR #42, opened 2019-07-18
Update to use Twisted and python 3.7

erikjohnston:


sygnal/apnspushkin.py
```diff
@@ -257,88 +392,28 @@ def get_payload_full(self, n):
...
-            payload['event_id'] = n.event_id
-
-        if n.counts.unread is not None:
-            payload['unread_count'] = n.counts.unread
-        if n.counts.missed_calls is not None:
-            payload['missed_calls'] = n.counts.missed_calls
-
-        return payload
+    async def shutdown(self):  # TODO
+        return await super().shutdown()
```
erikjohnston: Just remove this function, it will call super automatically.

sygnal/database.py
```diff
@@ -0,0 +1,77 @@
...
+import logging
+import sqlite3
+from concurrent.futures import ThreadPoolExecutor
+
+from twisted.internet.defer import Deferred
+
+logger = logging.getLogger(__name__)
+
+
+class Database:
```
erikjohnston:
```suggestion
class Database(object):
```

It's conventional to always derive from object. (In py2 if you don't then it uses different semantics)

sygnal/database.py
```diff
@@ -0,0 +1,77 @@
...
+    def __init__(self, dbfile, twisted_reactor):
+        self.twisted_reactor = twisted_reactor
+        self.dbfile = dbfile
+        # SQLite is blocking, so run queries in a separate thread
+        self.dbexecutor = ThreadPoolExecutor(max_workers=1)
```
erikjohnston: Might be worth looking at `aiosqlite`?

sygnal/sygnal.py
```diff
@@ -0,0 +1,318 @@
...
+        start_deferred.addCallback(on_started)
+        start_deferred.addErrback(on_failed_to_start)
+
+        logger.info("Starting pushkins")
+        self.reactor.run()
+
+        sys.exit(exit_code)
+
+    def shutdown(self):
+        pass  # TODO
```
erikjohnston: Pointless function is pointless :)

sygnal/sygnal.py
```diff
@@ -0,0 +1,318 @@
...
+            config (dict): Configuration for this Sygnal
+            custom_reactor: a Twisted Reactor to use.
+            tracer (optional): an OpenTracing tracer. The default is the no-op tracer.
+        """
+        self.config = config
+        self.reactor = custom_reactor
+        self.pushkins = {}
+        self.tracer = tracer
+
+    def _setup(self):
```
erikjohnston: Let's just fold this into init?

sygnal/gcmpushkin.py
```diff
@@ -47,197 +56,369 @@
...
+            "max_connections", DEFAULT_MAX_CONNECTIONS
+        )
+
+        self.http_agent = Agent(reactor=sygnal.reactor, pool=self.http_pool)
+
+        self.db = sygnal.database
+
+        logger.debug("About to set up CanonicalRegId Store")
+        self.canonical_reg_id_store = CanonicalRegIdStore()
+        await self.canonical_reg_id_store.setup(self.db)
```
erikjohnston: I'd rather we passed the store in via the constructor, that way a) `GcmPushkin` always has a valid store, b) we can then fold the start methods for the pushkins into the init as they no longer have to do async stuff which will make `run` a *lot* simpler.

sygnal/database.py
richvdh: > It's conventional to always derive from object

only if you want to support py2, imho

sygnal/gcmpushkin.py
reivilibre: This might be a struggle since the pushkin is constructed based on import name with a standard set of arguments that would be used to construct any pushkin. As such, passing it in the constructor would need adding special casing to ``.

We should also consider third-party pushkins which may benefit from being able to run async code at startup.

Thoughts? @erikjohnston

sygnal/database.py
reivilibre: AFAICT `aiosqlite` does the same thing (run all queries in serial on a worker thread) so I'm not sure if it would gain anything.

erikjohnston: Sure, but it saves me from having to figure out if your implementation is doing the right thing :)

sygnal/gcmpushkin.py
erikjohnston: > This might be a struggle since the pushkin is constructed based on import name with a standard set of arguments that would be used to construct any pushkin. As such, passing it in the constructor would need adding special casing for GCM.

We only have two types of pushkins so special casing is far from the end of world, there's something to be said for simplicity here. The other option is to add the store to the `Sygnal` object and then the `GcmPushkin` can grab the store from that. 

> We should also consider third-party pushkins which may benefit from being able to run async code at startup.

TBH dealing with third party pushkins is something to worry about when it happens.

--- 

Broadly I'm not a fan of having two steps to instansiate an object, it tends to lead to footguns where either the `setup` functions isn't called or is called more than once. Ideally once you have an instantiated object then it should be ready to be used, which basically means having to pass in everything the class needs into the constructor. If you really need some async setup then you probably want a static async method that does the work and then instantiates and returns the object.

sygnal/database.py
reivilibre: On another note, it would break testing since the test environment cannot use `asyncio` for async; pure Twisted only.

erikjohnston: Ah, then you probably want https://twistedmatrix.com/documents/current/core/howto/rdbms.html.

sygnal/gcmpushkin.py
reivilibre: I think my preferred solution then would be to have a static async method to construct everything. This gives the most flexibility without letting GCM-specific stuff (CanonicalRegIdStore) leech out of `gcmpushkin.py`.

I have already had 2 enquiries about custom pushkins, so it is probably worth offering the flexibility now. The code has been in place since last Sygnal, it may as well be left in.
Calling `async` code from a synchronous context can be a bit of a dog, I think it's worth leaving it open as an option. In the event that it isn't needed, nothing save from having to write 6 characters (`async `) is lost.

reivilibre:
> There are also a bunch of TODOs in the code that look like they either should be removed or dealt with :)

Regarding the TODOs:

* 2 of them correspond to waiting for changes in Twisted to separate from the global reactor – nothing can be done for now.
* 1 of them was in the previous Sygnal and might still be of interest (or not, I don't know)
* The others have gone.

sygnal/apnspushkin.py
```diff
@@ -13,240 +14,371 @@
...
+                return
+            prio = 10
+            if n.prio == "low":
+                prio = 5
+
+            shaved_payload = apnstruncate.truncate(
+                payload, max_length=self.MAX_JSON_BODY_SIZE
+            )
+
+            async def dispatch_request(span):
```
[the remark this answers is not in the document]
reivilibre: Do you mean pulling `dispatch_request` out of `dispatch_notification` or splitting `dispatch_request` somehow (if this – what do you suggest?)?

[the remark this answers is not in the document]
erikjohnston: Splitting it out of `dispatch_notification`

sygnal/gcmpushkin.py
```diff
@@ -47,197 +56,377 @@
...
+        return {reg_id: await self.get_canonical_id(reg_id) for reg_id in reg_ids}
 
-    def get_canonical_ids(self, reg_ids):
-        # TODO: Use one DB query
-        return {reg_id: self._get_canonical_id(reg_id) for reg_id in reg_ids}
+    async def get_canonical_id(self, reg_id):
+        rows = await self.db.runQuery(
+            "SELECT canonical_reg_id FROM gcm_canonical_reg_id WHERE reg_id = ?",
+            (reg_id,),
+        )
```
erikjohnston: Docstrings please :innocent:

PR #47, opened 2019-08-08
Isolate access logging and Twisted's logging

reivilibre:


sygnal/http.py
```diff
@@ -262,6 +262,21 @@ def count_deferred_code(_):
...
+    A subclass of Site to perform access logging in a way that makes sense for
+    Sygnal.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.logger = logging.getLogger("sygnal.access")
+
+    def log(self, request):
+        line = self._logFormatter(self._logDateTime, request) + u"\n"
```
erikjohnston: Let's not use private functions, as they may break from under us. In this case `self._logFormatter` is just `twisted.web.http.combinedLogFormatter` so we may as well use that directly.

(We also shouldn't need to add a newline)

reivilibre: `self._logFormatter` is populated from the `log_formatter` argument to `__init__`. Should I capture it directly instead? (Note that Sygnal changes formatter depending on circumstances)

erikjohnston: That also works

reivilibre: Also, for `_logDateTime`, should I copy the logic that maintains this and avoid referencing it?

erikjohnston: I wouldn't bother, we don't have enough traffic that `datetimeToLogString(self._reactor.seconds())` would be meaningfully slow tbh

reivilibre: Should I also capture the reactor myself?

PR #58, opened 2019-08-27
Handle unknown and 404 GCM response codes

reivilibre:
Fixes:
```
TypeError

cannot unpack non-iterable NoneType object
```

and throw exceptions to catch any other such instances in the future.

### DIFF
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -203,7 +203,7 @@ class GcmPushkin(Pushkin):
             )
             # permanent failure: give up
             raise NotificationDispatchException("Not authorised to push")
-        elif 200 <= response.code < 300:
+        elif 200 <= response.code < 300 or response.code == 404:
             try:
                 resp_object = json.loads(response_text)
             except JSONDecodeError:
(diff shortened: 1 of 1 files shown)


sygnal/gcmpushkin.py
```diff
@@ -203,7 +203,7 @@ def __init__(self, name, sygnal, config, canonical_reg_id_store):
             )
             # permanent failure: give up
             raise NotificationDispatchException("Not authorised to push")
-        elif 200 <= response.code < 300:
+        elif 200 <= response.code < 300 or response.code == 404:
```
erikjohnston: What does 404 mean? And do we really want to handle it the same as a success? If so then a comment about why this makes sense would be good

reivilibre: I'm not 100% sure – it's not documented.

However, [Stack Overflow](https://stackoverflow.com/questions/48928972/fcm-http-v1-api-returns-404-for-unregistered-token) answers suggest it will be used as the code when the device token is not registered.

Though looking more closely, it seems like perhaps it won't yield the same response format – so I guess we will treat it separately after all.

PR #62, opened 2019-09-23
Initial implementation of a Firebase Cloud Messaging pushkin

hawkowl:


sygnal/firebasepushkin.py
```diff
@@ -0,0 +1,263 @@
...
+        d = Deferred()
+
+        def done(success, result):
+            self.reactor.callFromThread(d.callback, result)
+
+        with SEND_TIME_HISTOGRAM.time():
+            self._pool.callInThreadWithCallback(
+                done, messaging.send_multicast, message, app=self._app
+            )
+            response = await d
```
erikjohnston: Can we not use `deferToThreadPool`?

sygnal/firebasepushkin.py
```diff
@@ -0,0 +1,263 @@
...
+
+        logger.debug(
+            "Message send success: %s of %s",
+            response.success_count,
+            response.success_count + response.failure_count,
+        )
+
+        failed = []
+
+        return failed
```
erikjohnston: Did we want to do something here rather than reporting success for everything? Did we want to catch exceptions from the send?

sygnal/firebasepushkin.py
```diff
@@ -0,0 +1,263 @@
...
+        failed = []
+        data = self.build_message(n)
+
+        logger.debug("Type: %s", data["type"])
+
+        if (
+            data["type"] != "m.room.message"
+            or data["content"]["msgtype"] not in self.config.message_types
+        ):
+            return failed
```
erikjohnston: So we're not sending push for non-message events? What about invites and things?

sygnal/firebasepushkin.py
```diff
@@ -0,0 +1,263 @@
...
+
+        unread_count = data["unread"] if data["unread"] is not None else 0
+        notification_title = (
+            data["sender_display_name"]
+            if data["room_name"] is None
+            else data["room_name"]
+        )
```
erikjohnston: This feels like something the apps should be doing?

sygnal/sygnal.py
```diff
@@ -159,14 +159,15 @@ def __init__(self, config, custom_reactor, tracer=opentracing.tracer):
             try:
                 self.pushkins[app_id] = await self._make_pushkin(app_id, app_cfg)
             except Exception:
-                logger.exception(
-                    "Failed to load and create pushkin for kind %s", app_cfg["type"]
+                raise Exception(
+                    "Failed to load and create pushkin for kind %s" % app_cfg["type"]
```
erikjohnston: Why?

setup.py
```diff
@@ -29,22 +29,31 @@ def read(fname):
...
+    "aioapns>=1.7",
+    "pyyaml>=5.1.1",
+    "service_identity>=18.1.0",
+    "zope.interface>=4.6.0",
+    "idna>=2.8",
+    "jaeger-client>=4.0.0",
+    "opentracing>=2.2.0",
+]
+
+EXTRAS = {"sentry": ["sentry-sdk>=0.10.2"], "firebase": ["firebase-admin>=3.0.0"]}
```
erikjohnston: Do we want to bother doing this? Doesn't it make it more annoying to install if you have to do `pip install sygnal[all]` to actually get support for firebase?

richvdh:
this seems to be bit-rotting

PR #76, opened 2019-12-10
Play default device sound upon receiving a push notification

anoadragon453:
Play the default device notification sound upon receiving an APNS push notification.

We should allow the user to specify what sound to play in Sygnal's config, but we can make things more feature-ful after they work :)


anoadragon453:
Eyetime have confirmed this is working for them, so merging.

PR #77, opened 2019-12-11
Improvements to the firebase implementation and adding per-event type handlers

anoadragon453:
Numerous fixes as part of the `jaywink/firebase-improvements` branch. Before merging, we should ensure everything within is useful to a wide audience.

From what I remember, a particular point that could be make configurable was the shape of the message contained within the notification. The implementation within the branch is currently quite specific.


richvdh:
this seems to be bit-rotting.

PR #87, opened 2020-03-23
Add prometheus metric for the number of requests in flight.

richvdh:
Based on #85

### DIFF
--- a/sygnal/http.py
+++ b/sygnal/http.py
@@ -14,27 +14,30 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import contextlib
 import json
 import logging
 import sys
 import time
 import traceback
+from typing import Optional
 from uuid import uuid4
 
-from opentracing import Format, tags, logs
-from prometheus_client import Counter, Histogram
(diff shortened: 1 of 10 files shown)


sygnal/http.py
```diff
@@ -164,9 +173,11 @@ def _handle_request(self, request):
 
             root_span_accounted_for = True
 
-            ensureDeferred(
-                self._handle_dispatch(root_span, request, log, notif, context)
-            )
+            async def cb():
+                with request.processing():
+                    await self._handle_dispatch(root_span, request, log, notif, context)
```
erikjohnston: Can we not just instead do:

```python
with REQUESTS_IN_FLIGHT_GUAGE.track_inprogress():
    await self._handle_dispatch(root_span, request, log, notif, context)
```

rather than having the request itself track what's going on? Or can we finish the request independently of the `_handle_dispatch`?

richvdh: well, the intention was to lay the groundwork for tracking requests in progress when we have other servlets which don't necessarily follow the same pattern. In principle, yes we can finish a request independently of the async handler - for example if we hand over control to a Producer to stream back a response.

In practice, we dont' do that anywhere in sygnal and honestly I guess it's unlikely that we ever will. So yeah, this is over-engineered. I'll simplify it.

PR #91, opened 2020-03-25
Support Postgresql

### LINKED ISSUE #89: Switch from sqlite to postgres
To allow for better performance and scalability for large deployments.

Half-Shot:
Fixes #89

### DIFF
--- a/sygnal.yaml.sample
+++ b/sygnal.yaml.sample
@@ -170,10 +170,9 @@ apps:
   #  #max_connections: 20
 
 db:
-  args:
-    host: localhost
-    database: sygnal
-    user: postgres
-    password: pass
-    cp_min: 1
-    cp_max: 5
+  host: localhost
+  database: sygnal
+  user: postgres
+  password: pass
+  cp_min: 1
+  cp_max: 5
--- a/sygnal.yaml.sample
+++ b/sygnal.yaml.sample
@@ -172,7 +172,7 @@ apps:
 db:
   host: localhost
   database: sygnal
-  user: postgres
+  user: sygnal
   password: pass
   cp_min: 1
   cp_max: 5
--- a/sygnal.yaml.sample
+++ b/sygnal.yaml.sample
@@ -3,20 +3,46 @@
 # See: matrix.org
 ##
 
-## Database #
-# You should uncomment one of these options to enable the appropriate database.
-# If host is provided, postgresql will be used. Otherwise sqlite will be selected
-# and will use the value of `dbfile` for it's databases.
-db:
-  ## for sqlite
-  dbfile: sygnal.db # Default
-  ## for postgresql
-  # host: localhost
-  # database: sygnal
-  # user: sygnal
-  # password: pass
-  # cp_min: 1
-  # cp_max: 5
+
+
+# The 'database' setting defines the database that sygnal uses to store all of
+# its data.
+#
+# 'name' gives the database engine to use: either 'sqlite3' (for SQLite) or
+# 'psycopg2' (for PostgreSQL).
+#
+# 'args' gives options which are passed through to the database engine,
+# except for options starting 'cp_', which are used to configure the Twisted
+# connection pool. For a reference to valid arguments, see:
+#   * for sqlite: https://docs.python.org/3/library/sqlite3.html#sqlite3.connect
+#   * for postgres: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS
+#   * for the connection pool: https://twistedmatrix.com/documents/current/api/twisted.enterprise.adbapi.ConnectionPool.html#__init__
+#
+#
+# Example SQLite configuration:
+#
+#database:
+#  name: sqlite3
+#  args:
+#    database: /path/to/database.db
+#
+#
+# Example Postgres configuration:
+#
+#database:
+#  name: psycopg2
+#  args:
+#    host: localhost
+#    database: sygnal
+#    user: sygnal
+#    password: pass
+#    cp_min: 1
+#    cp_max: 5
+#
+database:
+  name: sqlite3
+  args:
+    database: sygnal.db
 
 ## Logging #
 #
--- a/sygnal.yaml.sample
+++ b/sygnal.yaml.sample
@@ -3,6 +3,21 @@
 # See: matrix.org
 ##
 
+## Database #
+# You should uncomment one of these options to enable the appropriate database.
+# If host is provided, postgresql will be used. Otherwise sqlite will be selected
+# and will use the value of `dbfile` for it's databases.
+db:
+  ## for sqlite
+  dbfile: sygnal.db # Default
+  ## for postgresql
+  # host: localhost
+  # database: sygnal
+  # user: sygnal
+  # password: pass
+  # cp_min: 1
+  # cp_max: 5
+
 ## Logging #
 #
 log:
@@ -168,11 +183,3 @@ apps:
   #  # This is the maximum number of connections to GCM servers at any one time
   #  # the default is 20.
   #  #max_connections: 20
-
-db:
-  host: localhost
-  database: sygnal
-  user: sygnal
-  password: pass
-  cp_min: 1
-  cp_max: 5
--- a/sygnal.yaml.sample
+++ b/sygnal.yaml.sample
@@ -3,8 +3,6 @@
 # See: matrix.org
 ##
 
-
-
 # The 'database' setting defines the database that sygnal uses to store all of
 # its data.
 #
--- a/sygnal.yaml.sample
+++ b/sygnal.yaml.sample
@@ -169,3 +169,11 @@ apps:
   #  # the default is 20.
   #  #max_connections: 20
 
+db:
+  args:
+    host: localhost
+    database: sygnal
+    user: postgres
+    password: pass
+    cp_min: 1
+    cp_max: 5
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -452,6 +452,8 @@ class CanonicalRegIdStore(object):
 
         Args:
             db (adbapi.ConnectionPool): database to prepare
+            engine (str):
+                Database engine in use. This is either "sqlite" or "postgresql".
 
         """
         self.db = db
@@ -496,12 +498,12 @@ class CanonicalRegIdStore(object):
         rows = dict(
             await self.db.runQuery(
                 """
-            SELECT reg_id, canonical_reg_id
-            FROM gcm_canonical_reg_id
-            WHERE reg_id IN (%s)
-            """
+                SELECT reg_id, canonical_reg_id
+                FROM gcm_canonical_reg_id
+                WHERE reg_id IN (%s)
+                """
                 % (",".join(["?"] * len(reg_ids))),
                 reg_ids,
             )
         )
-        return {reg_id: dict(rows).get(reg_id, None) for reg_id in reg_ids}
+        return {reg_id: dict(rows).get(reg_id) for reg_id in reg_ids}
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -487,11 +487,13 @@ class CanonicalRegIdStore(object):
             or `None` if there is no entry.
         """
 
-        results = dict(await self.db.runQuery(
-            """SELECT reg_id, canonical_reg_id
+        results = dict(
+            await self.db.runQuery(
+                """SELECT reg_id, canonical_reg_id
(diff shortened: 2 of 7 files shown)


Half-Shot:
Unit tests won't pass without https://github.com/matrix-org/pipelines/pull/53

sygnal.yaml.sample
```diff
@@ -169,3 +169,10 @@ apps:
   #  # the default is 20.
   #  #max_connections: 20
 
+db:
+  host: localhost
+  database: sygnal
+  user: postgres
```
anoadragon453: Surely we shouldn't encourage using the `postgres` user for an application.

sygnal/gcmpushkin.py
```diff
@@ -482,23 +486,14 @@ def __init__(self):
...
-
-        Returns (dict):
-            its canonical registration ID, or `None` if there is no entry.
-        """
-        rows = await self.db.runQuery(
-            "SELECT canonical_reg_id FROM gcm_canonical_reg_id WHERE reg_id = ?",
-            (reg_id,),
+        results = dict(
+            await self.db.runQuery(
+                """SELECT reg_id, canonical_reg_id
```
anoadragon453: Can you make this prettier?

sygnal/sygnal.py
```diff
@@ -235,7 +237,7 @@ def check_section(section_name, known_keys, cfgpart=config):
     check_section(
         "access", {"file", "enabled", "x_forwarded_for"}, cfgpart=config["log"]
     )
-    check_section("db", {"dbfile"})
+    check_section("db", {"args"})
```
anoadragon453: Shouldn't be `args`.

tests/testutils.py
```diff
@@ -51,13 +95,14 @@ def setUp(self):
         )
 
         while not start_deferred.called:
-            # we need to advance until the pushkins have started up
+            # we need to wait for the database
             self.sygnal.reactor.advance(1)
             self.sygnal.reactor.wait_for_work(lambda: start_deferred.called)
```
anoadragon453: It looks like this is still waiting for `start_deferred`, which waits for the pushkins.

tests/testutils.py
```diff
@@ -24,15 +26,56 @@
...
+        conn = psycopg2.connect(
+            database=POSTGRES_DBNAME_FOR_INITIAL_CREATE,
+            user=POSTGRES_USER,
+            password=POSTGRES_PASSWORD,
+            host=POSTGRES_HOST,
+        )
+        conn.autocommit = True
+        cur = conn.cursor()
+        cur.execute("DROP DATABASE IF EXISTS %s;" % (dbname,))
+        cur.execute("CREATE DATABASE %s;" % (dbname,))
```
anoadragon453: You need to `conn.commit()` before closing.

tests/testutils.py
```diff
@@ -24,15 +26,56 @@
...
+    def _tear_down_database(self, dbname):
+        conn = psycopg2.connect(
+            database=POSTGRES_DBNAME_FOR_INITIAL_CREATE,
+            user=POSTGRES_USER,
+            password=POSTGRES_PASSWORD,
+            host=POSTGRES_HOST,
+        )
+        conn.autocommit = True
+        cur = conn.cursor()
+        cur.execute("DROP DATABASE %s;" % (dbname,))
```
anoadragon453: Same here.

sygnal/gcmpushkin.py
```diff
@@ -465,8 +465,12 @@ def __init__(self):
...
             canonical_reg_id (str): the canonical registration ID for `reg_id`
         """
-        await self.db.runQuery(
-            "INSERT OR REPLACE INTO gcm_canonical_reg_id VALUES (?, ?);",
+        await self.db.runOperation(
+            """
+            INSERT INTO gcm_canonical_reg_id VALUES (%s, %s)
+            ON CONFLICT (reg_id) DO UPDATE
+                 SET canonical_reg_id = EXCLUDED.canonical_reg_id;
+            """,
```
anoadragon453: Does `runOperation` `commit()` automatically or do we need to do so explicitly here as well?

sygnal/gcmpushkin.py
```diff
@@ -451,12 +451,12 @@ def __init__(self):
...
 
         Args:
-            db (adbapi.ConnectionPool): database to prepare
+            db (psycopg2.pool.AbstractConnectionPool): database to prepare
 
         """
         self.db = db
 
-        await self.db.runQuery(self.TABLE_CREATE_QUERY)
+        await self.db.runOperation(self.TABLE_CREATE_QUERY)
```
anoadragon453: Do we need to `commit()` here?

tests/testutils.py
Half-Shot: Not in this case. We're using `conn.autocommit` like in https://github.com/matrix-org/synapse/blob/d9239b52571c7e07fa9b035de1ad46bb602ed163/tests/utils.py#L68-L79 because running a DROP and a CREATE inside a single txn isn't allowed.

sygnal/gcmpushkin.py
Half-Shot: `runOperation` commits automatically

sygnal/gcmpushkin.py
Half-Shot: same as above.

tests/testutils.py
anoadragon453: Oh! Missed that, interesting

sygnal.yaml.sample
```diff
@@ -169,3 +169,10 @@ apps:
   #  # the default is 20.
   #  #max_connections: 20
 
+db:
```
richvdh: this should probably be nearer the top? and needs some comments. What does it do? what are the defaults if the settings are not given?

sygnal/sygnal.py
```diff
@@ -235,7 +237,6 @@ def check_section(section_name, known_keys, cfgpart=config):
     check_section(
         "access", {"file", "enabled", "x_forwarded_for"}, cfgpart=config["log"]
     )
-    check_section("db", {"dbfile"})
```
richvdh: does the updated section not merit any checks?

sygnal/gcmpushkin.py
```diff
@@ -91,7 +91,7 @@ class GcmPushkin(Pushkin):
     Pushkin that relays notifications to Google/Firebase Cloud Messaging.
     """
 
-    UNDERSTOOD_CONFIG_FIELDS = {"type", "api_key"}
+    UNDERSTOOD_CONFIG_FIELDS = {"type", "api_key", "max_connections"}
```
richvdh: is this related?

anoadragon453:
We assumed since we assume that nobody else runs this, switching all deployments to postgres wouldn't be an issue.

I know we can't *know* we aren't the only ones running this, but that's what we've assumed in the past.

sygnal/sygnal.py
Half-Shot: I've added some checks below. It's a bit *ugh* because we are checking for different things depending on the database being used, but it's better than nothing.

sygnal.yaml.sample
```diff
@@ -3,6 +3,21 @@
 # See: matrix.org
 ##
 
+## Database #
```
richvdh: could you make this follow the synapse conventions for sample configs? I feel like it's been serving us well there.

richvdh:
> We assumed since we assume that nobody else runs this, switching all deployments to postgres wouldn't be an issue.

A few points that occur to me here. Not to pick on either of you guys, but just some things I think we all need to think about as we try to get ourselves into a more mature situation where we aren't always doing the bare minimum to put out the current fire:

* As you say, we can't assume that we are the only people running this, and the fact that we have done so in the past (and presumably totally messed other people about in the process) isn't a great reason to do so now.
* Even if we *were* the only people running this, I'd be keen that we maintain support for sqlite for a while at least. Operationally, things are much easier if you can separate software upgrades from configuration changes: it means you don't have to change everything at once (and, often more importantly, you don't have to change everything back when it doesn't work.)

Similarly, I'm keen that we document stuff like this clearly, even if we don't expect other people to run it. The reasons for this are:

 * New Vector is now big enough that the people writing the code may or may not be the people deploying it, or - particularly - the people fixing it at 2am. When I've just been woken up to fix a problem, clear documentation is a godsend.
 * Even if we get this deployed and configured and it all runs perfectly, there will come a point in 6 or 12 months when we need to re-deploy it or something. And at that point, we'll have forgotten everything we knew. Rather than have to figure it out again from first principles, we could look at the documentation we wrote while it was fresh in our heads.

In both cases, imagining that you are writing for an external audience will be a great exercise in making sure that we have good documentation, even if that external audience doesn't really exist.

sygnal/sygnal.py
Half-Shot: I've since added checks that `name` and `args` are present.

sygnal.yaml.sample
Half-Shot: Have copied across

sygnal/gcmpushkin.py
```diff
@@ -443,7 +443,7 @@ class CanonicalRegIdStore(object):
     def __init__(self):
         self.db = None
 
-    async def setup(self, db):
+    async def setup(self, db, engine):
```
richvdh: docstring for `engine` please

sygnal/gcmpushkin.py
```diff
@@ -482,23 +493,15 @@ def __init__(self):
...
-        rows = await self.db.runQuery(
-            "SELECT canonical_reg_id FROM gcm_canonical_reg_id WHERE reg_id = ?",
-            (reg_id,),
+        rows = dict(
+            await self.db.runQuery(
+                """
+            SELECT reg_id, canonical_reg_id
+            FROM gcm_canonical_reg_id
+            WHERE reg_id IN (%s)
+            """
```
richvdh: this lot could do with more indentation, for consistency with other `"""` blocks

sygnal/gcmpushkin.py
```diff
@@ -482,23 +493,15 @@ def __init__(self):
             mapping of registration ID to either its canonical registration ID,
             or `None` if there is no entry.
         """
-        return {reg_id: await self.get_canonical_id(reg_id) for reg_id in reg_ids}
```
richvdh: not saying it's a bad idea, but... is this actually part of supporting postgres or is it an unrelated change?

sygnal/sygnal.py
```diff
@@ -76,6 +76,17 @@ def __init__(self, config, custom_reactor, tracer=opentracing.tracer):
         observer = twisted_log.PythonLoggingObserver()
         observer.start()
 
+        # Old format db config
+        if config.get("db", None) is not None:
```
richvdh: `dict.get` defaults to None, and it's pythonic to skip the explicit `is not None`

```suggestion
        if config.get("db"):
```

sygnal/sygnal.py
```diff
@@ -76,6 +76,17 @@ def __init__(self, config, custom_reactor, tracer=opentracing.tracer):
...
+            logger.warning(
+                "Config includes the legacy 'db' option, please migrate"
+                + "to 'database' instead"
```
richvdh: "migrateto" is two words. `+` operator is redundant.

```suggestion
                " to 'database' instead"
```

sygnal.yaml.sample
```diff
@@ -3,6 +3,45 @@
 # See: matrix.org
 ##
 
+# The 'database' setting defines the database that sygnal uses to store all of
```
richvdh: this documentation is nice now. :+1:

sygnal/gcmpushkin.py
richvdh: (for clarity: I'm not saying you need to change it back now, but for future ref, assuming it is unrelated, I'd prefer it to be in a separate PR.)

Half-Shot: Yep, I got a bit eager with changing something that seemed like a performance problem. Luckily there are unit tests testing this path. In the future, I will open a separate PR :+1:

sygnal/sygnal.py
Half-Shot: Good to know :+1:

sygnal/gcmpushkin.py
```diff
@@ -452,11 +452,14 @@ def __init__(self):
 
         Args:
             db (adbapi.ConnectionPool): database to prepare
+            engine (str):
+                Database engine in use. This is either "sqlite" or "postgresql".
```
richvdh: it's not in use yet?

```suggestion
                Database engine to use. Shoud be either "sqlite" or "postgresql".
```

PR #95, opened 2020-04-01
Fix postgres database configuration breakage

Half-Shot:
This fixes:
- The default sqlite3 configuration from colliding with PostgreSQL
- Tests not testing ^
- Postgresql not actually working properly due to '?' '%s' confusion in the query.

### DIFF
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -495,7 +495,7 @@ class CanonicalRegIdStore(object):
             mapping of registration ID to either its canonical registration ID,
             or `None` if there is no entry.
         """
-        parameterKey = "?" if self.engine == "sqlite" else "%s"
+        parameter_key = "?" if self.engine == "sqlite" else "%s"
         rows = dict(
             await self.db.runQuery(
                 """
@@ -503,7 +503,7 @@ class CanonicalRegIdStore(object):
                 FROM gcm_canonical_reg_id
                 WHERE reg_id IN (%s)
                 """
-                % (",".join(parameterKey for _ in reg_ids)),
+                % (",".join(parameter_key for _ in reg_ids)),
                 reg_ids,
             )
         )
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -495,6 +495,7 @@ class CanonicalRegIdStore(object):
             mapping of registration ID to either its canonical registration ID,
             or `None` if there is no entry.
         """
+        parameterKey = "?" if self.engine == "sqlite" else "%s"
         rows = dict(
             await self.db.runQuery(
                 """
@@ -502,7 +503,7 @@ class CanonicalRegIdStore(object):
                 FROM gcm_canonical_reg_id
                 WHERE reg_id IN (%s)
                 """
-                % (",".join("%s" for _ in reg_ids)),
+                % (",".join(parameterKey for _ in reg_ids)),
                 reg_ids,
             )
         )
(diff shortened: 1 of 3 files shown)


sygnal/gcmpushkin.py
```diff
@@ -495,14 +495,15 @@ def __init__(self):
...
+        parameterKey = "?" if self.engine == "sqlite" else "%s"
         rows = dict(
             await self.db.runQuery(
                 """
                 SELECT reg_id, canonical_reg_id
                 FROM gcm_canonical_reg_id
                 WHERE reg_id IN (%s)
                 """
-                % (",".join(["?"] * len(reg_ids))),
+                % (",".join(parameterKey for _ in reg_ids)),
```
anoadragon453, line 506:
```suggestion
                % (",".join(parameter_key for _ in reg_ids)),
```

sygnal/gcmpushkin.py
```diff
@@ -495,14 +495,15 @@ def __init__(self):
             mapping of registration ID to either its canonical registration ID,
             or `None` if there is no entry.
         """
+        parameterKey = "?" if self.engine == "sqlite" else "%s"
```
anoadragon453, line 498:
```suggestion
        parameter_key = "?" if self.engine == "sqlite" else "%s"
```

sygnal/sygnal.py
```diff
@@ -86,6 +85,11 @@ def __init__(self, config, custom_reactor, tracer=opentracing.tracer):
                 "name": "sqlite3",
                 "args": {"dbfile": config["db"]["dbfile"]},
             }
+        elif config.get("database") is None:
+            config["database"] = {
+                "name": "sqlite3",
+                "args": {"dbfile": "sygnal.db"},
+            }
```
anoadragon453, line 92: How is this different from what you removed?

Half-Shot: `merge_left_with_defaults` (in the `if __name__ == '__main__'`) statement would do a deep merge and  copy subkeys if the parent is present, which would cause `dbfile` to appear even if `args` was specified.

The change ensures `database` is only defined if it's not already defined.

tests/testutils.py
```diff
@@ -78,20 +78,20 @@ def _tear_down_database(self, dbname):
...
         cur.close()
         conn.close()
 
     def setUp(self):
         reactor = ExtendedMemoryReactorClock()
 
         config = {"apps": {}, "log": {"setup": {"version": 1}}}
-        config = merge_left_with_defaults(CONFIG_DEFAULTS, config)
 
         self.config_setup(config)
```
anoadragon453, line 92: Not sure if you want this line as it looks like it'll try and add `dbfile` regardless of whether you're using postgres or sqlite?

sygnal/sygnal.py
anoadragon453: Right, makes sense.

tests/testutils.py
Half-Shot: Are you sure? I can't see that?

anoadragon453: https://github.com/matrix-org/sygnal/blob/b784e46f3a22700efac8560d0ab0230e7c3f955f/tests/testutils.py#L34 ?

anoadragon453: Ah, I was using an old checkout. This method looks fine on master :)

richvdh:
shouldn't this have had a changelog file?

Half-Shot:
Yes, I forgot that and included one in the release :|. To pay for my sins, I'll set up a newsfile CI

richvdh:
thanks. If you'd included the newsfile here, we could have pointed out that the word "crash" is misleading...

Half-Shot:
fair point

PR #97, opened 2020-04-03
Fix the given sample config to match the implementation

csett86:
The implementation for sqlite3 expects a dbfile, so match the sample
provided to that.


Half-Shot:
Thanks for the contribution. Would you mind signing off and adding a changelog file as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md ?

csett86:
@Half-Shot, thank you for pointing out the contribution guidelines. I have amended the pull request, it should be good now. If there is anthing else to add, please just let me know.

PR #100, opened 2020-04-06
Allow the 'db' config section to be defined

### LINKED ISSUE #98: Config checker doesn't allow for deprecated fields
E.g. we removed `db` as an option, which now means sygnal will not start with `db`.

We should probably make the config checker more relaxed about unknown fields.

Half-Shot:
Fixes #98


clokep:
This seems to work well in my testing.

PR #106, opened 2020-04-22
Report the certificate expiration as a prometheus metric.

### LINKED ISSUE #103: Certificate expiry monitoring
Sygnal should check the expiry date of its certificates and emit them as a metric which can trigger alerts.

clokep:
This reports the epoch time (in seconds) of the APNs certificate expiry as a prometheus metric. The metric is updated each time the certificate is loaded (so each time sygnal is started).

pyOpenSSL becomes a direct dependency instead of just a transient one via aioapns.

Fixes #103

### DIFF
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -148,7 +148,7 @@ class ApnsPushkin(Pushkin):
 
         def report_expiry():
             seconds_left = int((expiration_date - datetime.utcnow()).total_seconds())
-            CERTIFICATE_EXPIRATION_GAUGE.set(seconds_left)
+            CERTIFICATE_EXPIRATION_GAUGE.labels(pushkin=self.name).set(seconds_left)
 
         # Report the metric every 60 seconds.
         looper = LoopingCall(report_expiry)
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -54,7 +54,7 @@ RESPONSE_STATUS_CODES_COUNTER = Counter(
 )
 
 CERTIFICATE_EXPIRATION_GAUGE = Gauge(
-    "sygnal_cert_expiry",
+    "sygnal_client_cert_expiry",
     "The expiry date of the client certificate in seconds since the epoch",
     labelnames=["pushkin"],
 )
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -27,6 +27,7 @@ from opentracing import logs, tags
 import OpenSSL
 from prometheus_client import Histogram, Counter, Gauge
 from twisted.internet.defer import Deferred
+from twisted.internet.task import LoopingCall
 
 from sygnal import apnstruncate
 from sygnal.exceptions import (
@@ -125,7 +126,6 @@ class ApnsPushkin(Pushkin):
                 client_cert=self._certfile, use_sandbox=self.use_sandbox
             )
 
-            # TODO Should this continually report.
             self._report_certificate_expiration()
         else:
             self.apns_client = APNs(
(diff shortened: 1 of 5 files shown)


clokep:
I think this is ready, but leaving as a draft until I get confirmation that seconds-until-expiration is the proper way to report this.

sygnal/apnspushkin.py
```diff
@@ -51,6 +53,12 @@
     labelnames=["pushkin", "code"],
 )
 
+CERTIFICATE_EXPIRATION_GAUGE = Gauge(
+    "sygnal_cert_expiry",
```
clokep, line 57: This name could probably use bikeshedding, let me know if there's a particular pattern we use.

clokep:
Updated this based on discussion on the issue. This should be ready to go now!

sygnal/apnspushkin.py
richvdh: be good to identify it's a client cert rather than the server cert. maybe `sygnal_client_cert_expiry`.

sygnal/apnspushkin.py
```diff
@@ -128,6 +136,20 @@ def __init__(self, name, sygnal, config):
...
+    def _report_certificate_expiration(self, certfile):
+        """Export the epoch time that the certificate expires as a metric."""
+        with open(certfile, "rb") as f:
+            cert = f.read()
+
+        x509 = OpenSSL.crypto.load_certificate(OpenSSL.crypto.FILETYPE_PEM, cert)
+        # Convert from a string to a datetime object.
+        expiration_date = datetime.strptime(
+            x509.get_notAfter().decode(), "%Y%m%d%H%M%SZ"
+        )
```
richvdh, line 148: my impression (possibly unfounded) is that generally it's preferred to use `pyca/cryptography` rather than doing battle with openssl directly. https://cryptography.io/en/latest/x509/reference/#loading-certificates makes it look fairly easy. WDYT?

sygnal/apnspushkin.py
```diff
@@ -128,6 +136,20 @@ def __init__(self, name, sygnal, config):
...
+        # Convert the expiration time to seconds since the epoch.
+        epoch_time = int((expiration_date - datetime(1970, 1, 1)).total_seconds())
```
richvdh, line 150: /me stares at this a bit. does it do the the right thing with respect to timezones?

Also, can we not  just use `expiration_date.timestamp()`?

richvdh: oh, the incantation is possibly a workaround for `expiration_date` not having correct timezone info? I guess it would be helpful to write that as a comment.

clokep: Ah, I grabbed this from another project which had Python 2.7 compatibility still. Didn't know `timestamp()` existed to be honest.

sygnal/apnspushkin.py
clokep: If that's preferred we can do it that way, looks like it is pretty much the same.

For the record I used OpenSSL directly here since this is what aioapns uses (and I had code which did most of what I needed already).

clokep: Much to my surprise.... pyOpenSSL requires cryptography...

clokep:
Thanks @richvdh! I made some updates from your feedback. The timezone handling is a bit weird, but [cryptography says](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Certificate.not_valid_after):

> A naïve datetime representing the end of the validity period for the certificate in UTC. This value is inclusive.

And the [`timestamp()`](https://docs.python.org/3.7/library/datetime.html#datetime.datetime.timestamp) docs say:

> **Note**:
> There is no method to obtain the POSIX timestamp directly from a naive `datetime` instance representing UTC time. If your application uses this convention and your system timezone is not set to UTC, you can obtain the POSIX timestamp by supplying `tzinfo=timezone.utc`:
>
> `timestamp = dt.replace(tzinfo=timezone.utc).timestamp()`
> 
> or by calculating the timestamp directly:
>
> `timestamp = (dt - datetime(1970, 1, 1)) / timedelta(seconds=1)`

I believe I've combined these two pieces of info correctly! :)

PR #108, opened 2020-04-23
Reuse Configurations With Asterisk App IDs and Token Based APNS Auth

ismailgulek:
This PR aims to enable having appId's like _im.vector.*_ in sygnal.yaml file, which will enable using the same authentication to APNS against multiple appId's.
It also aims to enable token-based authentication for APNS, which is the only way to reuse authentication when notifying multiple apps.

### DIFF
--- a/sygnal/http.py
+++ b/sygnal/http.py
@@ -205,7 +205,7 @@ class V1NotifyHandler(Resource):
     def find_pushkins(self, appid):
         """
         Finds matching pushkins in self.sygnal.pushkins according to the appid.
-        If finds a specific pushkin with the exact app id, immediately returns
+        If it finds a specific pushkin with the exact app id, immediately returns
         it (as a list).
         Otherwise returns possible pushkins (as a list).
 
--- a/sygnal/http.py
+++ b/sygnal/http.py
@@ -205,8 +205,10 @@ class V1NotifyHandler(Resource):
     def find_pushkins(self, appid):
         """Finds matching pushkins in self.sygnal.pushkins according to the appid.
 
+
         Args:
             appid (str): app identifier to search in self.sygnal.pushkins.
+
         Returns:
             list of `Pushkin`: If it finds a specific pushkin with
                 the exact app id, immediately returns it.
--- a/sygnal/http.py
+++ b/sygnal/http.py
@@ -205,7 +205,8 @@ class V1NotifyHandler(Resource):
     def find_pushkins(self, appid):
         """
         Finds matching pushkins in self.sygnal.pushkins according to the appid.
-        If finds a specific pushkin with the exact app id, immediately returns it (as a list).
+        If finds a specific pushkin with the exact app id, immediately returns
+        it (as a list).
         Otherwise returns possible pushkins (as a list).
 
         appid: app identifier to search in self.sygnal.pushkins.
--- a/sygnal/http.py
+++ b/sygnal/http.py
@@ -203,15 +203,15 @@ class V1NotifyHandler(Resource):
                 root_span.finish()
 
     def find_pushkins(self, appid):
-        """
-        Finds matching pushkins in self.sygnal.pushkins according to the appid.
-        If it finds a specific pushkin with the exact app id, immediately returns
-        it (as a list).
-        Otherwise returns possible pushkins (as a list).
+        """Finds matching pushkins in self.sygnal.pushkins according to the appid.
 
-        appid: app identifier to search in self.sygnal.pushkins.
+        Args:
+            appid (str): app identifier to search in self.sygnal.pushkins.
+        Returns:
+            list of `Pushkin`: If it finds a specific pushkin with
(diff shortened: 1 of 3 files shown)


sygnal/http.py
```diff
@@ -201,6 +202,28 @@ async def cb():
             if not root_span_accounted_for:
                 root_span.finish()
 
+    def find_pushkins(self, appid):
+        """
+        Finds matching pushkins in self.sygnal.pushkins according to the appid.
+        If finds a specific pushkin with the exact app id, immediately returns
```
babolivier, line 208:
```suggestion
        If it finds a specific pushkin with the exact app id, immediately returns
```

sygnal/http.py
```diff
@@ -201,6 +202,28 @@ async def cb():
...
+        it (as a list).
+        Otherwise returns possible pushkins (as a list).
+
+        appid: app identifier to search in self.sygnal.pushkins.
```
babolivier, line 212: We're usually using Google's docstring style for arguments and such: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html (e.g. https://github.com/matrix-org/synapse/blob/54283f3ed49ce40059e9f0746f2ded08ce3a5b9e/synapse/util/threepids.py#L22-L32)

What this means is that arguments should look like:

```
Args:
    arg1 (type1): short description
    arg2 (type2): short description
```

We haven't always been good at always remembering to mention the type, and it's not mandatory, but most Python IDEs support it and it helps them to show accurate docs and do some better type checking/linting. Could you rewrite that part of the docstring to match this style please?

sygnal/http.py
```diff
@@ -201,6 +202,28 @@ async def cb():
...
+        if appid in self.sygnal.pushkins:
+            return [self.sygnal.pushkins[appid]]
+
+        result = []
+        for key, value in self.sygnal.pushkins.items():
+            # The ".+" symbol is used in place of "*" symbol
+            regex = key.replace("*", ".+")
+            if re.search(regex, appid):
+                result.append(value)
+        return result
```
babolivier, line 225: Please document the return value in the docstring along with its type.

tests/test_http.py
```diff
@@ -0,0 +1,123 @@
...
+
+        # Assert
+        # method should be called one time
+        self.assertEqual(1, method.call_count)
+
+        self.assertEqual({"rejected": []}, resp)
+
+    def test_with_matching_appid(self):
+        """
+        Tests the matching case: A matching app (only one time) id must be processed.
```
babolivier, line 92:
```suggestion
        Tests the matching case: A matching app id (only one time) must be processed.
```

"app id" is the name of something here, it sounds wrong to split it.

tests/test_http.py
```diff
@@ -0,0 +1,123 @@
...
+DEVICE_EXAMPLE_AMBIGIOUS = {
+    "app_id": "com.example.apns2",
+    "pushkey": "spqr",
+    "pushkey_ts": 42,
+}
+
+
+class HttpTestCase(testutils.TestCase):
+    def setUp(self):
+        self.apns_mock_class = patch("sygnal.apnspushkin.APNs").start()
```
babolivier, line 51: Given `APNs` is only present in `sygnal.apnspushkin` as an import from `aioapns`, why not use `aioapns.APNs` here?

richvdh:
> However, I feel like I'm missing a lot of context on what issue does this solve in the first place and why do we need this patch. Could you please provide us with some details about that?

+1. please don't just send in PRs without any description whatsoever about what they are supposed to do and why.

sygnal/http.py
ismailgulek: Sure 👍

tests/test_http.py
ismailgulek: Hmm, it wasn't a conscious choice. I just copied it from another test class. Updated.

sygnal/http.py
```diff
@@ -203,15 +203,15 @@ async def cb():
...
-        Finds matching pushkins in self.sygnal.pushkins according to the appid.
-        If it finds a specific pushkin with the exact app id, immediately returns
-        it (as a list).
-        Otherwise returns possible pushkins (as a list).
+        """Finds matching pushkins in self.sygnal.pushkins according to the appid.
 
-        appid: app identifier to search in self.sygnal.pushkins.
+        Args:
+            appid (str): app identifier to search in self.sygnal.pushkins.
+        Returns:
```
babolivier, line 210:
```suggestion

        Args:
            appid (str): app identifier to search in self.sygnal.pushkins.

        Returns:
```

(sections should be separated by blank lines)

tests/test_http.py
ismailgulek: Well, i could not make it work with `aioapns`. Tests were failed. Had to revert.

richvdh:
ahhhh, "asterisk" as in wildcard, as opposed to "Asterisk" as in the popular telephony software...

Issue #109, opened 2020-04-24
The following configuration fields are not understood: {'database'}


theodotos:
I am running sygnal following the instructions in the README: `python -m sygnal.sygnal`

I am using sqlite3 exactly like in the sample configuration:

```
database:
  name: sqlite3
  args:
    dbfile: sygnal.db
```

But I get this:

```
The following configuration fields are not understood: {'database'}
```

Any idea?

clokep:
@theodotos What version of Sygnal are you running?

theodotos:
@clokep 

I cloned it recently from git:

```
# git log | head
commit 8567224dd5112599cbe0a997aaccc075d7b5c254
Merge: c8a1216 1f33ce1
Author: Will Hunt <will@half-shot.uk>
Date:   Mon Apr 6 21:41:49 2020 +0100

    Merge pull request #97 from csett86/fix-sample-config
    
    Fix the given sample config to match the implementation

commit 1f33ce1f3208602b173db28eba0142c4281d828e
```

theodotos:
@clokep 

I pulled the latest version:

```
# git log | head
commit 44aa835086af0567061d26a47dd10a6aec912ba9
Author: ismailgulek <gulekismail@gmail.com>
Date:   Fri Apr 24 16:56:50 2020 +0300

    Update changelog
    
    Signed-off-by: ismailgulek <gulekismail@gmail.com>

commit dc3ecc3172a4937eb2ac586c20b40a7e38950b18
Author: ismailgulek <gulekismail@gmail.com>
```

Same:

```
# python -m sygnal.sygnal
The following configuration fields are not understood: {'database'}

```

babolivier:
Hey there, and thanks for the report! I've opened https://github.com/matrix-org/sygnal/pull/111 to fix this issue.

babolivier:
Turns out #100 is already fixing this.

theodotos:
I didn't mention that the service runs, when this happens but does not listen to port 5000.

babolivier:
For what it's worth, this log line is only a warning and shouldn't impact Sygnal's ability to start and run properly (it's much more of a logging mess than an actual bug). I think your port 5000 issue might be unrelated.

anilk9096:
hi ,

when can we expect this fix , with out sygnal server we cant migrate our app due to push notifications .

any alternate for sygnal . kindly suggest

richvdh:
fixed by #100 I believe

PR #111, opened 2020-04-28
Move default database config to CONFIG_DEFAULTS

babolivier:
Otherwise https://github.com/matrix-org/sygnal/issues/109 happens, because it seems that, to Sygnal, "not understood" = "doesn't have defaults" (and doesn't mean the config will be ignored). This PR fixes https://github.com/matrix-org/sygnal/issues/109 but we probably should think about making this log clearer, if it's needed at all.

### DIFF
--- a/sygnal/sygnal.py
+++ b/sygnal/sygnal.py
@@ -50,6 +50,7 @@ CONFIG_DEFAULTS = {
         "sentry": {"enabled": False},
     },
     "apps": {},
+    "database": {"name": "sqlite3", "args": {"dbfile": "sygnal.db"}},
 }
 
 
@@ -84,11 +85,6 @@ class Sygnal(object):
             config["database"] = {
                 "name": "sqlite3",
                 "args": {"dbfile": config["db"]["dbfile"]},
-            }
-        elif config.get("database") is None:
-            config["database"] = {
-                "name": "sqlite3",
-                "args": {"dbfile": "sygnal.db"},
             }
 
         sentrycfg = config["metrics"]["sentry"]
@@ -293,6 +289,14 @@ def merge_left_with_defaults(defaults, loaded_config):
 
     # copy defaults or override them
     for k, v in result.items():
+        if k == "database" and loaded_config.get("database"):
+            # If the provided configuration includes a database section, ignore
+            # defaults.
+            # This is because the args sub-section doesn't have the same attributes
+            # depending on the database engine, and the engine might complain if given
+            # an attribute it doesn't understand.
+            continue
+
         if isinstance(v, dict):
             if k in loaded_config:
                 result[k] = merge_left_with_defaults(v, loaded_config[k])


sygnal/sygnal.py
```diff
@@ -50,6 +50,7 @@
         "sentry": {"enabled": False},
     },
     "apps": {},
+    "database": {"name": "sqlite3", "args": {"dbfile": "sygnal.db"}},
```
anoadragon453, line 53: The opposite of this PR's diff was purposefully done in https://github.com/matrix-org/sygnal/pull/95/files#diff-13c51378d37f89582bc66a30f28f723b So I'm afraid this code won't work.

We either need to make `merge_left_with_defaults` more sane, or consider removing the log line as you say.

babolivier: I went for the former in https://github.com/matrix-org/sygnal/pull/111/commits/526dd8ae215c1fb238e71aab14e460deaa015a10, which probably makes more sense and is easier to do. Sorry about the force-push, I hadn't seen your review.

anoadragon453: I feel that rather than signalling out the database here, a better fix might just be to pull out the specific args Postgres needs rather than passing the entire config dict:

https://github.com/matrix-org/sygnal/blob/6a70885421d27fecf5145cf3d20fd0811fa37680/sygnal/sygnal.py#L139-L141

as that seems to be what the issue is here (postgres complaining about getting a `dbfile` argument?

babolivier: It is that issue indeed. But I feel like doing it that way would not win us anything and make it more painful in terms of maintenance because the current way of doing it also how we do it in all of our other projects (so creating an exception to that can easily make debugging issues more difficult - not saying we shouldn't do it, but rather that we should maintain consistency between our projects), and if an update to psycopg2 that changes the way a connection pool is configured it would require both a code change and a config change, whereas the current solution only requires the latter.

anoadragon453: I suppose that makes sense, although it's worth mentioning we already hard-code the options for SQLite3. Using `**args` was something introduced only recently with the postgres PRs.

To me it feels a bit weird to special-case the database here, but the comment explains its reasoning so in the grand scheme of things it's probably fine.

richvdh:
I'd like to have a look at this before we merge it. It feels like we've fixed the same thing 3 times recently.

sygnal/sygnal.py
anoadragon453: Actually, testing with `ConnectionPool` it doesn't complain if you give it `dbfile` as an arg, it just ignores it.

anoadragon453:
Ok, so I talked with @Half-Shot and he's already written a fix for this in https://github.com/matrix-org/sygnal/pull/100, where `"database"` is set to `None` in `CONFIG_DEFAULTS`, which prevents the warning.

Though in terms of fixing https://github.com/matrix-org/sygnal/issues/109, it seems like something we should get rid of. We ignore extra config options in our other projects.

sygnal/sygnal.py
babolivier: > Actually, testing with ConnectionPool it doesn't complain if you give it dbfile as an arg with "psycopg2", it just ignores it.

Well tests were failing with errors complaining explicitly about a `dbfile` parameter in the arguments so...

> Using `**args` was something introduced only recently with the postgres PRs.

In Sygnal yes, but surely we've been doing that in Sydent and Synapse for a while now? I definitely got that error before on Synapse due to a `path` in the args in a bad config file.

babolivier:
> Ok, so I talked with @Half-Shot and he's already written a fix for this in #100, where "database" is set to None in CONFIG_DEFAULTS, which prevents the warning.

Oh actually that might be a better way to do it. Though I'd still find it weird that we have two different defaults at two different places for the same thing.

> Though in terms of fixing #109, it seems like something we should get rid of. We ignore extra config options in our other projects.

Agreed, but I wouldn't get rid of it right away - I'd assume it was added for a reason and I wouldn't remove it without knowing what that reason is and what ops think about it.

richvdh:
> Ok, so I talked with @Half-Shot and he's already written a fix for this in #100, where "database" is set to None in CONFIG_DEFAULTS, which prevents the warning.

right. I *thought* I'd already seen a fix to this recently :)

richvdh:
sounds like we're favouring #100 then?

babolivier:
Yes probably.

PR #114, opened 2020-05-06
Push Payload Changes

ismailgulek:
Change `content-available` to `mutable-content` and also add `event_id` in payload if provided. This is intended to enable modifying push notification content on the client side (iOS). For details:

https://github.com/vector-im/riot-ios/issues/2714

### DIFF
--- a/changelog.d/114.feature
+++ b/changelog.d/114.feature
@@ -1,0 +1,1 @@
+Change aps payload to be mutable, include event_id in payload.
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -426,7 +426,7 @@ class ApnsPushkin(Pushkin):
             aps["badge"] = badge
 
         if loc_key:
-            aps["content-available"] = 1
+            aps["mutable-content"] = 1
 
         if loc_key is None and badge is None:
             log.info("Nothing to do for alert of type %s", n.type)
@@ -436,6 +436,8 @@ class ApnsPushkin(Pushkin):
 
         if loc_key and n.room_id:
             payload["room_id"] = n.room_id
+        if loc_key and n.event_id:
+            payload["event_id"] = n.event_id
 
         payload["aps"] = aps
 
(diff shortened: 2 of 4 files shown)


ismailgulek:
@babolivier thanks for the review. I'll try to explain the change as far as i can.

`content-available` key is used to make a notification silent, when the key is set to 1, `alert`, `badge` and `sound` keys should be omitted. Detailed info about silent notifications is [here](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/pushing_background_updates_to_your_app).

`mutable-content` key is used to make a notification modifiable in client side. We will use this mechanism as a replacement of VoIP pushes and our intention is to modify notification content (e.g. decrypt event) and if we fail on this, we'll still have some generic content to show (e.g. "Somebody sent you a message"). Detailed info about modifying notification content is [here](https://developer.apple.com/documentation/usernotifications/modifying_content_in_newly_delivered_notifications).

Actually our usage of `content-available` key was useless because we were also setting the `alert` and `badge` keys. So these notifications were not silent. Indeed, we were not using regular notifications at all, but VoIP pushes.

Detailed info about payload keys can be found [here](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification#2943363).

babolivier:
@ismailgulek Right, so neither `content-available` nor `content-available` are actually passed on to clients?

ismailgulek:
> @ismailgulek Right, so neither `content-available` nor `content-available` are actually passed on to clients?

No, they are both passed. But they don't help us at the moment, because we're using VoIP pushes. They will be important with our new development.

babolivier:
> No, they are both passed

Right, my point is that Riot isn't the only mobile client that's using push so we should care about backwards compatibility in case one actually does something with these parameters.

ismailgulek:
> > No, they are both passed
> 
> Right, my point is that Riot isn't the only mobile client that's using push so we should care about backwards compatibility in case one actually does something with these parameters.

It should not affect any other clients, as they already should be using VoIP pushes.

PR #122, opened 2020-05-12
Log errors during start-up and fix the default logging config

### LINKED ISSUE #120: The sample logging configuration doesn't work
If used as is, the sample logging configuration makes Sygnal not log anything, neither in stdout, stderr nor sygnal.log.

When tweaking it to try to make it work somehow, I also couldn't get

https://github.com/matrix-org/sygnal/blob/9a1e3c8acddf8464e228b97aa2756f0bd3ba0848/sygnal/sygnal.py#L190-L192

or

https://github.com/matrix-org/sygnal/blob/9a1e3c8acddf8464e228b97aa2756f0bd3ba0848/sygnal/sygnal.py#L196-L196

or any log line that didn't originate from Twisted, to log, both in sygnal.log and in stderr.

### LINKED ISSUE #121: sys.exit(1) doesn't actually stop Sygnal
In `_make_pushkins_then_start`, we call `sys.exit(1)` when failing to create pushkins:

https://github.com/matrix-org/sygnal/blob/9a1e3c8acddf8464e228b97aa2756f0bd3ba0848/sygnal/sygnal.py#L185-L197

Instead of causing Sygnal to shut down, they instead get caught by Twisted, which errors but doesn't stop the reactor:

```
2020-05-12 12:51:20,826 [16441] CRITICAL twisted Unhandled error in Deferred:
2020-05-12 12:51:20,827 [16441] CRITICAL twisted 
  File "/home/brendan/.virtualenvs/sygnal-py38/lib/python3.8/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/home/brendan/Documents/matrix/sygnal/sygnal/sygnal.py", line 193, in _make_pushkins_then_start
    sys.exit(1)
SystemExit: 1
```

clokep:
This copies (a slimmed down) [version of the code we use for Synapse](https://github.com/matrix-org/synapse/blob/207b1737ee0acd226359d59ce3b7f7d46111b1c8/synapse/app/homeserver.py#L395-L423) to log errors during start-up.

With this I'm able to get a sane error message if I have a bad config:

```
$ python -m sygnal.sygnal
The following configuration fields are not understood: {'database'}
Error during startup:
  File "/.virtualenvs/sygnal/lib/python3.7/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/sygnal/sygnal/sygnal.py", line 218, in start
    port, bind_addresses, pushgateway_api
  File "/.virtualenvs/sygnal/lib/python3.7/site-packages/twisted/internet/defer.py", line 911, in ensureDeferred
    return _cancellableInlineCallbacks(coro)
  File "/.virtualenvs/sygnal/lib/python3.7/site-packages/twisted/internet/defer.py", line 1529, in _cancellableInlineCallbacks
    _inlineCallbacks(None, g, status)
--- <exception caught here> ---
  File "/sygnal/sygnal/sygnal.py", line 218, in start
    port, bind_addresses, pushgateway_api
  File "/.virtualenvs/sygnal/lib/python3.7/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/sygnal/sygnal/sygnal.py", line 196, in _make_pushkins_then_start
    "No app IDs are configured. Edit sygnal.yaml to define some."
builtins.RuntimeError: No app IDs are configured. Edit sygnal.yaml to define some.
```

Fixes #121
Fixes #120

### DIFF
--- a/sygnal/sygnal.py
+++ b/sygnal/sygnal.py
@@ -28,9 +28,10 @@ import prometheus_client
 import yaml
 from opentracing.scope_managers.asyncio import AsyncioScopeManager
 from twisted.enterprise.adbapi import ConnectionPool
-from twisted.internet import asyncioreactor
+from twisted.internet import asyncioreactor, defer
 from twisted.internet.defer import ensureDeferred
 from twisted.python import log as twisted_log
+from twisted.python.failure import Failure
 
 from sygnal.http import PushGatewayApiServer
 
@@ -130,10 +131,9 @@ class Sygnal(object):
                     )
                     raise
             else:
-                logger.error(
+                raise RuntimeError(
                     "Unknown OpenTracing implementation: %s.", tracecfg["impl"]
                 )
-                sys.exit(1)
 
         db_name = config["database"]["name"]
 
@@ -187,14 +187,14 @@ class Sygnal(object):
             try:
                 self.pushkins[app_id] = await self._make_pushkin(app_id, app_cfg)
             except Exception:
-                logger.exception(
+                raise RuntimeError(
                     "Failed to load and create pushkin for kind %s", app_cfg["type"]
                 )
-                sys.exit(1)
 
         if len(self.pushkins) == 0:
-            logger.error("No app IDs are configured. Edit sygnal.yaml to define some.")
-            sys.exit(1)
+            raise RuntimeError(
+                "No app IDs are configured. Edit sygnal.yaml to define some."
+            )
 
         logger.info("Configured with app IDs: %r", self.pushkins.keys())
 
@@ -210,9 +210,25 @@ class Sygnal(object):
         bind_addresses = self.config["http"]["bind_addresses"]
         pushgateway_api = PushGatewayApiServer(self)
 
-        ensureDeferred(
-            self._make_pushkins_then_start(port, bind_addresses, pushgateway_api)
-        )
+        @defer.inlineCallbacks
+        def start():
+            try:
+                yield ensureDeferred(
+                    self._make_pushkins_then_start(
+                        port, bind_addresses, pushgateway_api
+                    )
+                )
+            except Exception:
+                # Print the exception and bail out.
+                print("Error during startup:", file=sys.stderr)
+
+                # this gives better tracebacks than traceback.print_exc()
+                Failure().printTraceback(file=sys.stderr)
+
+                if self.reactor.running:
+                    self.reactor.stop()
+
+        self.reactor.callWhenRunning(start)
         self.reactor.run()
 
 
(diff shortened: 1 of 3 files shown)


sygnal/sygnal.py
```diff
@@ -210,9 +210,25 @@ def run(self):
         bind_addresses = self.config["http"]["bind_addresses"]
         pushgateway_api = PushGatewayApiServer(self)
 
-        ensureDeferred(
```
clokep, line 213: This in particular had me scratch my head since...it makes a `Deferred` but never does anything with it.

clokep:
One thing to note with this is that the process does exit with a success code (`0`). Adding a `sys.exit(1)` caused an additional exception to be raised, but still exited as success.

sygnal/sygnal.py
```diff
@@ -210,9 +210,25 @@ def run(self):
...
+        def start():
+            try:
+                yield ensureDeferred(
+                    self._make_pushkins_then_start(
+                        port, bind_addresses, pushgateway_api
+                    )
+                )
+            except Exception:
+                # Print the exception and bail out.
+                print("Error during startup:", file=sys.stderr)
```
babolivier, line 223: Shouldn't this be `logger.error` or something similar? Or maybe we're assuming that logging doesn't work because of https://github.com/matrix-org/sygnal/issues/120, so we're only printing for now?

sygnal/sygnal.py
```diff
@@ -210,9 +210,25 @@ def run(self):
...
+
+                # this gives better tracebacks than traceback.print_exc()
+                Failure().printTraceback(file=sys.stderr)
```
babolivier, line 226: Same as above.

sygnal/sygnal.py
clokep: You cannot guarantee that logging has been configured yet. At least I assume that's why the Synapse code does this.

babolivier: Why? Logging is configured in the `Sygnal` constructor, so it has to have been configured when calling `sygnal.run`.

clokep: It is a pattern I've seen before, I think it guards against:
* Logging not being configured
* Logging being configured in a broken way (e.g. #120)

If I use `logger.error` here I see nothing in the logs, but this is likely #120 as you pointed out.

clokep:
I also added the fix for #120 to this.

sygnal/sygnal.py
babolivier: sgtm

sygnal/sygnal.py
reivilibre: iirc that schedules that coroutine on the Reactor (if I remember the Twisted words correctly).

PR #127, opened 2020-06-10
Use `default_payload` in Push Payloads

ismailgulek:
Add `fallback_content` case handling from push data to include `alert` and `mutable-content` fields in the `aps` dictionary.

Part of vector-im/riot-ios#3325

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -65,6 +65,16 @@ gcm
   to contain the 'Server key', which can be acquired from Firebase Console at:
   ``https://console.firebase.google.com/project/<PROJECT NAME>/settings/cloudmessaging/``
 
+PusherData Configurations
+=======
+
+When calling https://matrix.org/docs/spec/client_server/r0.5.0#post-matrix-client-r0-pushers-set, clients
+can provide a ``default_payload`` data dictionary in ``PusherData``, which will be passed to Sygnal.
+This will be used by Sygnal to have a default data dictionary in the push payload, for both APNS and
+GCM pushers, if provided. Upon the default payload dictionary, Sygnal will make incremental changes.
+This is useful for clients to decide default push payload content. For instance, iOS clients will have
+freedom to use silent/mutable notifications and be able to set some default alert/sound/badge fields.
+
 Running
 =======
 
--- a/README.rst
+++ b/README.rst
@@ -69,7 +69,7 @@ Pusher `data` configuration
 =======
 
 The following parameters can be specified in the `data` dictionary which is given when configuring the pusher
-via [POST /_matrix/client/r0/pushers/set](<https://matrix.org/docs/spec/client_server/latest#post-matrix-client-r0-pushers-set>) :
+via `POST /_matrix/client/r0/pushers/set <https://matrix.org/docs/spec/client_server/latest#post-matrix-client-r0-pushers-set>`_ :
 
 * ``default_payload``: a dictionary which defines the basic payload to be sent to the notification service.
   Sygnal will merge information specific to the push event into this dictionary. If unset, the empty dictionary is used.
--- a/README.rst
+++ b/README.rst
@@ -68,12 +68,14 @@ gcm
 Pusher `data` configuration
 =======
 
-When calling https://matrix.org/docs/spec/client_server/r0.5.0#post-matrix-client-r0-pushers-set, clients
-can provide a ``default_payload`` data dictionary in ``PusherData``, which will be passed to Sygnal.
-This will be used by Sygnal to have a default data dictionary in the push payload, for both APNS and
-GCM pushers, if provided. Upon the default payload dictionary, Sygnal will make incremental changes.
-This is useful for clients to decide default push payload content. For instance, iOS clients will have
-freedom to use silent/mutable notifications and be able to set some default alert/sound/badge fields.
+The following parameters can be specified in the `data` dictionary which is given when configuring the pusher
+via [POST /_matrix/client/r0/pushers/set](<https://matrix.org/docs/spec/client_server/latest#post-matrix-client-r0-pushers-set>) :
+
+* ``default_payload``: a dictionary which defines the basic payload to be sent to the notification service.
+  Sygnal will merge information specific to the push event into this dictionary. If unset, the empty dictionary is used.
+
+  This can be useful for clients to specify default push payload content. For instance, iOS clients will have
+  freedom to use silent/mutable notifications and be able to set some default alert/sound/badge fields.
 
 Running
 =======
--- a/README.rst
+++ b/README.rst
@@ -65,7 +65,7 @@ gcm
   to contain the 'Server key', which can be acquired from Firebase Console at:
   ``https://console.firebase.google.com/project/<PROJECT NAME>/settings/cloudmessaging/``
 
-PusherData Configurations
+Pusher `data` configuration
 =======
 
 When calling https://matrix.org/docs/spec/client_server/r0.5.0#post-matrix-client-r0-pushers-set, clients
--- a/changelog.d/127.feature
+++ b/changelog.d/127.feature
@@ -1,1 +1,1 @@
-Add fallback_content handling from push data.
+Add fallback_content check to include alert field in the APNS payload.
--- a/changelog.d/127.feature
+++ b/changelog.d/127.feature
@@ -1,0 +1,1 @@
+Add fallback_content handling from push data.
--- a/changelog.d/127.feature
+++ b/changelog.d/127.feature
@@ -1,1 +1,1 @@
-Add fallback_content check to include alert field in the APNS payload.
+Use default_payload from device data for both APNS and GCM payloads.
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -430,7 +430,7 @@ class ApnsPushkin(Pushkin):
         payload = {}
 
         if n.type and device.data:
-            payload = copy.deepcopy(device.data.get("default_payload",{}))
+            payload = copy.deepcopy(device.data.get("default_payload", {}))
 
         payload.setdefault("aps", {})
 
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -428,7 +428,11 @@ class ApnsPushkin(Pushkin):
 
         payload = {}
 
-        if device.data is not None and device.data["default_payload"] is not None:
+        if (
+            n.type is not None
+            and device.data is not None
+            and device.data["default_payload"] is not None
+        ):
             payload = device.data["default_payload"]
 
         if "aps" not in payload:
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -434,9 +434,7 @@ class ApnsPushkin(Pushkin):
         payload.setdefault("aps", {})
 
         if loc_key:
-            if "alert" not in payload["aps"]:
-                payload["aps"]["alert"] = {}
-            payload["aps"]["alert"]["loc-key"] = loc_key
+            payload["aps"].setdefault("alert", {})["loc-key"] = loc_key
 
         if loc_args:
             if "alert" not in payload["aps"]:
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -228,7 +228,7 @@ class ApnsPushkin(Pushkin):
             if n.event_id and not n.type:
                 payload = self._get_payload_event_id_only(n, device)
             else:
-                payload = self._get_payload_full(n, log)
+                payload = self._get_payload_full(n, device, log)
 
             if payload is None:
                 # Nothing to do
@@ -291,6 +291,9 @@ class ApnsPushkin(Pushkin):
         """
         payload = {}
 
+        if device.data is not None and device.data["default_payload"] is not None:
+            payload = device.data["default_payload"]
+
         if n.room_id:
             payload["room_id"] = n.room_id
         if n.event_id:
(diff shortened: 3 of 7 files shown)


sygnal/apnspushkin.py
```diff
@@ -299,6 +299,13 @@ def _get_payload_event_id_only(self, n):
         if n.counts.missed_calls is not None:
             payload["missed_calls"] = n.counts.missed_calls
 
+        payload["aps"] = {"mutable-content": 1}
```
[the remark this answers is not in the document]
ismailgulek: Yes, it's standard. It can be seen at https://developer.apple.com/documentation/usernotifications/modifying_content_in_newly_delivered_notifications

sygnal/apnspushkin.py
```diff
@@ -299,6 +299,13 @@ def _get_payload_event_id_only(self, n):
...
+        if (
+            n.devices[0].data["fallback_content"] is not None
+            and n.devices[0].data["fallback_content"]
```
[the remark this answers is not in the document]
ismailgulek: I'm not sure about that, the `fallback_content` will be an additional field in `data` property of https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-pushers-set

sygnal/apnspushkin.py
```diff
@@ -299,6 +299,13 @@ def _get_payload_event_id_only(self, n):
         if n.counts.missed_calls is not None:
             payload["missed_calls"] = n.counts.missed_calls
 
+        payload["aps"] = {"mutable-content": 1}
+        if (
+            n.devices[0].data["fallback_content"] is not None
```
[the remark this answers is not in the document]
ismailgulek: You're right. Updated.

changelog.d/127.feature
```diff
@@ -0,0 +1 @@
+Add fallback_content handling from push data.
```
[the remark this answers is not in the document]
ismailgulek: Sure, `fallback_content` is needed because we need to add `alert` in the `aps` field of payload. The handling is this: if we see this value as _true_ in the data, we're adding `alert` to the payload. It's because of this: 

> The system executes your notification content app extension only when a remote notification’s payload contains the following information:
> - The payload must include the mutable-content key with a value of 1.
> - The payload must include an alert dictionary with title, subtitle, or body information.

Detailed info can be found at https://developer.apple.com/documentation/usernotifications/modifying_content_in_newly_delivered_notifications 

It only affects iOS and does not affect Android or other pusher types.

sygnal/apnspushkin.py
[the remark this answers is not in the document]
richvdh: that seems to suggest that we need to include "an alert dictionary with title, subtitle, or body information" if we set `mutable-content` ?

sygnal/apnspushkin.py
```diff
@@ -278,7 +278,7 @@ async def dispatch_notification(self, n, device, context):
                             retry_delay, twisted_reactor=self.sygnal.reactor
                         )
 
-    def _get_payload_event_id_only(self, n):
+    def _get_payload_event_id_only(self, n, device):
```
richvdh, line 281: `device` needs an entry in the docstring, please (preferably one that says what type it is)

sygnal/apnspushkin.py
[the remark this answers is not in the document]
richvdh: The thing is, for client implementations to know what to set, and for pushgateway implementations to know how to interpret it, it has to be specced, otherwise this becomes secret knowledge which will only work with riot-ios and sygnal.

sygnal/apnspushkin.py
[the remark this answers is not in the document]
ismailgulek: Yes, and we're using a localization dictionary (`{"loc-key": "SINGLE_UNREAD", "loc-args": []}`) as `title`, which is also supported.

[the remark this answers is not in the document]
richvdh: ok, I guess I'll take your word for this. thanks for the clarification!

sygnal/apnspushkin.py
ismailgulek: updated

ismailgulek:
Decided to update this with a more general solution.

Offering use of `default_payload` in `data` which will be some JSON data, set by the clients. This will be even more helpful because in this way clients will have more freedom to use mutable/silent or any other type of notifications, and also have possibility to select `alert` string or dictionary in the way they want. These are some configurations for iOS side, but also Android side can have some default payload.

sygnal/apnspushkin.py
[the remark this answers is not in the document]
richvdh: as per https://github.com/matrix-org/matrix-doc/pull/2631#issuecomment-646535932: no, it doesn't need to be in the spec. Sorry!

That said, I *do* think it would be useful if we documented somewhere within the sygnal project what parameters the `apnspushkin` supports in its `data` dict and what they do. Perhaps you could add these details under https://github.com/matrix-org/sygnal#app-types ?

sygnal/apnspushkin.py
```diff
@@ -278,17 +278,22 @@ async def dispatch_notification(self, n, device, context):
...
+            device (Device): Device information to which the constructed payload
+            will be sent.
 
         Returns:
             The APNs payload as a nested dicts.
         """
         payload = {}
 
+        if device.data is not None and device.data["default_payload"] is not None:
+            payload = device.data["default_payload"]
```
richvdh, line 295: I think this will fail if there is no `default_payload` in `device.data`.

Also, I think you need to make sure you copy the `default_payload` dict here, otherwise any updates you do to it in the next few lines will overwrite the data on the `Device`.

```suggestion
        if device.data:
            payload.update(device.data.get("default_payload", {}))
```

sygnal/apnspushkin.py
```diff
@@ -422,25 +422,40 @@ def _get_payload_full(self, n, log):
...
             return None
 
         payload = {}
 
+        if (
+            n.type is not None
+            and device.data is not None
+            and device.data["default_payload"] is not None
+        ):
+            payload = device.data["default_payload"]
```
richvdh, line 436: same again, but because we now have multiple levels of dict, we'll have to do a proper [deepcopy](https://docs.python.org/3/library/copy.html#copy.deepcopy):

```suggestion
        if n.type and device.data:
            payload = copy.deepcopy(device.data.get("default_payload",{}))
```

sygnal/apnspushkin.py
```diff
@@ -422,25 +422,40 @@ def _get_payload_full(self, n, log):
...
+
+        if "aps" not in payload:
+            payload["aps"] = {}
```
richvdh, line 439: you can write this more succinctly:

```suggestion
        payload.setdefault("aps", {})
```

sygnal/apnspushkin.py
```diff
@@ -422,25 +422,40 @@ def _get_payload_full(self, n, log):
...
+
+        if loc_key:
+            if "alert" not in payload["aps"]:
+                payload["aps"]["alert"] = {}
+            payload["aps"]["alert"]["loc-key"] = loc_key
```
richvdh, line 444:
```suggestion
            payload["aps"].setdefault("alert", {})["loc-key"] = loc_key
```

sygnal/apnspushkin.py
```diff
@@ -422,25 +422,40 @@ def _get_payload_full(self, n, log):
                 badge = 0
             badge += n.counts.missed_calls
 
-        if badge is not None:
-            aps["badge"] = badge
-
-        if loc_key:
-            aps["mutable-content"] = 1
```
richvdh, line 429: is removing this deliberate?

sygnal/gcmpushkin.py
```diff
@@ -393,16 +393,22 @@ async def dispatch_notification(self, n, device, context):
...
+            device (Device): Device information to which the constructed payload
+            will be sent.
 
         Returns:
             JSON-compatible dict
         """
         data = {}
+
+        if device.data is not None and device.data["default_payload"] is not None:
+            data = device.data["default_payload"]
```
richvdh, line 410:
```suggestion
        if device.data:
            data.update(device.data.get("default_payload", {}))
```

sygnal/apnspushkin.py
richvdh: likewise below

sygnal/apnspushkin.py
ismailgulek: > I think this will fail if there is no `default_payload` in `device.data`.

Actually it doesn't. There are tests handling this. But copying makes sense.

sygnal/apnspushkin.py
ismailgulek: Yes sure, clients will decide contents of the `aps` with `default_payload` anymore.

ismailgulek:
> A couple of things to fix here.
> 
> Please can you add some documentation for each of the two pushkins, which documents the structure of the `data` field they expect?

Sure, is it ok to add these in Pushkin classes, like: https://github.com/matrix-org/sygnal/blob/d63e70f163df7d4e068b3c8d1c07134b380a5701/sygnal/apnspushkin.py#L66

sygnal/apnspushkin.py
richvdh: I don't see any tests for this...

ismailgulek: There are some tests without a `data.default_payload`, which was enough for this check. Like: https://github.com/matrix-org/sygnal/blob/d63e70f163df7d4e068b3c8d1c07134b380a5701/tests/test_apns.py#L26

richvdh: that has no `data` either. the problem came when there is a `data` but no `default_payload`.

ismailgulek: Ah sorry, you're right 👍

ismailgulek:
> code looks good now. I look forward to seeing the docs!

thanks, README updated

README.rst
```diff
@@ -65,6 +65,16 @@ gcm
...
+PusherData Configurations
+=======
+
+When calling https://matrix.org/docs/spec/client_server/r0.5.0#post-matrix-client-r0-pushers-set, clients
+can provide a ``default_payload`` data dictionary in ``PusherData``, which will be passed to Sygnal.
+This will be used by Sygnal to have a default data dictionary in the push payload, for both APNS and
+GCM pushers, if provided. Upon the default payload dictionary, Sygnal will make incremental changes.
+This is useful for clients to decide default push payload content. For instance, iOS clients will have
+freedom to use silent/mutable notifications and be able to set some default alert/sound/badge fields.
+
```
richvdh, line 77: I think it would be clearer to start a list here.

```
The following parameters can be specified in the `data` dictionary which is given when configuring the pusher
via `POST /_matrix/client/r0/pushers/set <https://matrix.org/docs/spec/client_server/latest#post-matrix-client-r0-pushers-set>`_ :

* ``default_payload``: a dictionary which defines the basic payload to be sent to the notification service. 
  Sygnal will merge information specific to the push event into this dictionary. If unset, the empty dictionary is used.

  This can be useful for clients to specify default push payload content.... etc
```

ismailgulek: Thanks for the suggestion, updated

README.rst
```diff
@@ -65,6 +65,18 @@ gcm
   to contain the 'Server key', which can be acquired from Firebase Console at:
   ``https://console.firebase.google.com/project/<PROJECT NAME>/settings/cloudmessaging/``
 
+Pusher `data` configuration
+=======
+
+The following parameters can be specified in the `data` dictionary which is given when configuring the pusher
+via [POST /_matrix/client/r0/pushers/set](<https://matrix.org/docs/spec/client_server/latest#post-matrix-client-r0-pushers-set>) :
```
richvdh, line 72: this is RST, so this format doesn't work. I think I got it right in my suggestion?

ismailgulek: Ah sorry yes, you're right. I was bitten by my previewer. Updated.

ismailgulek:
> looks great, thank you for bearing with it!

Great, thanks for your valuable comments.

myoussef89:
I got this error when I send it empty, what shoud be sent inside ?:

Jun 28 22:08:54 ip-172-31-11-241 python3[4208]: 2020-06-28 22:08:54,462 [4208] ERROR sygnal.http [40efe488-af22-41ee-8fc2-da59be7238b7] Exception whilst dispatching notification.
Jun 28 22:08:54 ip-172-31-11-241 python3[4208]: Traceback (most recent call last):
Jun 28 22:08:54 ip-172-31-11-241 python3[4208]:   File "/home/ubuntu/sygnal/sygnal/http.py", line 265, in _handle_dispatch
Jun 28 22:08:54 ip-172-31-11-241 python3[4208]:     result = await pushkin.dispatch_notification(notif, d, context)
Jun 28 22:08:54 ip-172-31-11-241 python3[4208]:   File "/home/ubuntu/sygnal/sygnal/gcmpushkin.py", line 326, in dispatch_notification
Jun 28 22:08:54 ip-172-31-11-241 python3[4208]:     data = GcmPushkin._build_data(n, device)
Jun 28 22:08:54 ip-172-31-11-241 python3[4208]:   File "/home/ubuntu/sygnal/sygnal/gcmpushkin.py", line 410, in _build_data
Jun 28 22:08:54 ip-172-31-11-241 python3[4208]:     data.update(device.data.get("default_payload", {}))
Jun 28 22:08:54 ip-172-31-11-241 python3[4208]: ValueError: dictionary update sequence element #0 has length 1; 2 is required

"data": {
                "default_payload": {},
                "format": "event_id_only",
                "url": "https://matrix.friends-groups.com/_matrix/push/v1/notify"
            }

ismailgulek:
> I got this error when I send it empty, what shoud be sent inside ?:

Nothing actually. I've tested your data dictionary and it was fine in my Sygnal. I'm not sure what's the root cause but you may remove `default_payload` field.

PR #130, opened 2020-06-30
WIP: HTTP CONNECT proxy support

reivilibre:
This PR adds support for proxying outbound traffic (to APNs and FCM) through a HTTP proxy using the CONNECT method.

It also includes support for HTTP proxy basic authentication.


reivilibre:
I will note that ~400 lines of this ~800 line PR have been pulled from Synapse (the Twisted_ stuff) with minor modifications to accept proxy basic auth.

sygnal/apnspushkin.py
```diff
@@ -454,3 +489,101 @@ async def _send_notification(self, request):
...
+
+
+class ProxyingEventLoopWrapper:
+    """
+    This is a wrapper for an asyncio.AbstractEventLoop which intercepts calls to
+    create_connection and transparently tunnels them through an HTTP CONNECT
+    proxy.
+
+    Note that this may only work against aioapns and perhaps not in a wider
+    setting.
```
clokep, line 501: Is this worth trying to upstream to `aioapns`?

clokep: Ah never-mind, I see that this uses a bunch of other internal classes.

sygnal.yaml.sample
```diff
@@ -123,6 +123,24 @@ http:
...
+  # Specify an address and port to use for connecting to the HTTP proxy.
+  #
+  address: '127.0.0.1:8080'
+
+  # If your proxy uses HTTP proxy basic authentication, specify the username
+  # and password here.
+  #
+  #username:
+  #password:
+
```
richvdh, line 143: I don't strongly oppose this, but is there a reason to prefer it over the semi-standard `https_proxy` env var?

reivilibre: Good point.

On one hand, you might want the Sygnal config all in one place. (I'd have to use a shell script to launch Sygnal with a proxy then...)

On the other hand, you might want the system's config in one place...

I don't know if there are security implications of putting secrets in an env var? Feels slightly dodgy but that's probably just because I wouldn't at the command-line...

I'll trust whatever you think is best but that's my take on it.

richvdh: @michaelkaye : do you have any opinions on this?

michaelkaye: So I've definitely seen a lot of situations where the proxy is run with the envvar in place - it's not something particularly weird to see.

The only time it might not be a good idea is if you have multiple proxy configurations for different pieces of the system, eg one for talking to google and one for talking to apple.

But practically, one http_proxy envvar is fine.

michaelkaye: it would be good to take the format of the http proxy envvar as well; 

http://user:pass@host:port/

reivilibre: Thanks! I've put this in now

sygnal/apnspushkin.py
```diff
@@ -131,8 +137,36 @@ def __init__(self, name, sygnal, config):
...
+                if proxycfg.get("username") and proxycfg.get("password"):
+                    logger.info("Using HTTP Proxy Basic Auth")
+                    http_proxy_creds = (proxycfg["username"], proxycfg["password"])
+                else:
+                    logger.info("No HTTP Proxy credentials configured")
+
+        loop = asyncio.get_event_loop()
+        if http_proxy_url:
+            # this overrides the create_connection method to use a HTTP proxy
+            loop = ProxyingEventLoopWrapper(loop, http_proxy_url, http_proxy_creds)
```
richvdh, line 164: ProxyingEventLoopWrapper appears to expect a host:port rather than a URL?

(suggests more tests are needed?)

sygnal/apnspushkin.py
```diff
@@ -455,3 +490,101 @@ async def _send_notification(self, request):
...
+        proxy_url = urlparse(self.proxy_address)
+
+        # MUST await on this to receive TLS alerts
+        # in the event of no TLS, it will be marked as complete unconditionally
+        tls_waiter = Future()
+
+        def make_protocol():
+            if ssl:
+
+                def make_ssl():
```
richvdh, line 545: I can't help but feel that nested functions inside nested functions is at least one level too much nesting...

how about an `HttpConnectProtocolFactory` class which you can instantiate here and then pass in `factory.make_protocol` into `create_connection` ?

sygnal/apnspushkin.py
```diff
@@ -455,3 +490,101 @@ async def _send_notification(self, request):
...
+        self._wrapped_loop = wrapped_loop
+        self.proxy_address = proxy_address
+        self.proxy_basic_auth = proxy_basic_auth
+
+    async def create_connection(
+        self,
+        protocol_factory: Callable[[], asyncio.Protocol],
+        host: str,
+        port: int,
+        ssl=False,
```
richvdh, line 534: what's the type of `ssl` here?
```suggestion
        ssl: Union[bool, SSLContext]=False,
```

sygnal/apnspushkin.py
```diff
@@ -455,3 +490,101 @@ async def _send_notification(self, request):
...
+                    top_protocol = protocol_factory()
+                    ssl_protocol = SSLProtocol(
+                        loop=self._wrapped_loop,
+                        app_protocol=top_protocol,
+                        sslcontext=context,
+                        waiter=tls_waiter,
+                        server_side=False,
+                        server_hostname=proxy_url.hostname,
+                    )
+                    return ssl_protocol, top_protocol
```
richvdh, line 559: this seems to use private bits of the asyncio package...

aren't you supposed to use https://docs.python.org/3/library/asyncio-eventloop.html#tls-upgrade?

sygnal/gcmpushkin.py
```diff
@@ -113,10 +114,35 @@ def __init__(self, name, sygnal, config, canonical_reg_id_store):
...
 
-        self.http_agent = Agent(
+        http_proxy_url = None
+        http_proxy_creds = None
+
+        # use the Sygnal global proxy configuration
+        proxycfg = sygnal.config["proxy"]
+
+        if proxycfg.get("enabled", False):
+            http_proxy_url = proxycfg.get("address")
```
richvdh, line 124: is this a url or a host:port?

sygnal/gcmpushkin.py
```diff
@@ -389,6 +417,10 @@ async def dispatch_notification(self, n, device, context):
 
             if len(pushkeys) > 0:
                 log.info("Gave up retrying reg IDs: %r", pushkeys)
+                if len(pushkeys) == total_pushkey_count:
+                    # if every single device failed, we should properly declare
+                    # this a failure
+                    raise NotificationDispatchException("Gave up retrying FCM.")
```
richvdh, line 423: is this a separate thing?

sygnal/helper/asyncio_connectprotocol.py
```diff
@@ -0,0 +1,178 @@
...
+                protocol of the stack (even if it is an unfortunately ugly
+                design).
+
+                Pseudo-example:
+                ```
+                def factory():
+                    http = HttpProtocol(...)
+                    ssl = SSLProtocol(app_protocol=http, ...)
+                    return (ssl, http)
+                ```
```
richvdh, line 76: yay for clear docs, but suggest taking some of this essay outside the `args` list!

reivilibre:
Thanks for your thorough review! I'll get through to it shortly.

Yes, it's really ugly to have two implementations of the same thing.
In the past, I found trying to bridge asyncio and Twisted was quite messy, so I can't say I tried too hard this time before giving in to just using 2 implementations. If you think it's worthwhile sharing, I can have a go.

sygnal/sygnal.py
```diff
@@ -77,6 +79,32 @@ def __init__(self, config, custom_reactor, tracer=opentracing.tracer):
         observer = twisted_log.PythonLoggingObserver()
         observer.start()
 
+        proxy_config = config["proxy"]
+        if proxy_config["enabled"].lower() == "env":
+            # Neither casing seems 100% ubiquitous so we will accept either
+            logger.info("Will use proxy configuration from environment if available.")
+            proxy_url_upper = os.getenv("HTTPS_PROXY")
+            proxy_url_lower = os.getenv("https_proxy")
```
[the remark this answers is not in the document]
reivilibre: https://curl.haxx.se/libcurl/c/libcurl-env.html

curl will do both, except for `HTTP_PROXY` where it only accepts `http_proxy`.

A few cases of each are scattered across the place, though in general, users are probably used to setting both upper- and lower-case proxy variables…

So I suppose we could just go back to expecting the idiomatic `HTTPS_PROXY`. Shall I do this?

README.rst
```diff
@@ -77,6 +77,20 @@ via `POST /_matrix/client/r0/pushers/set <https://matrix.org/docs/spec/client_se
   This can be useful for clients to specify default push payload content. For instance, iOS clients will have
   freedom to use silent/mutable notifications and be able to set some default alert/sound/badge fields.
 
+Using an HTTP Proxy for outbound traffic
+----------------------------------------
+
+Sygnal will, by default, automatically detect a ``HTTPS_PROXY`` or ``HTTP_PROXY``
```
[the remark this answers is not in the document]
reivilibre: oops, was going to do both cases but yes, let's stick with uppercase in the README at least.

sygnal.yaml.sample
```diff
@@ -123,6 +123,31 @@ http:
...
+  # Currently only HTTP proxies with CONNECT capability are supported.
+  # Specify:
+  # - `true` to use the settings here
+  # - `env` to automatically detect the settings from the de facto standard
+  #   `HTTPS_PROXY`/`https_proxy` environment variables.
+  # - `false` to be strict about not using one, not even from the environment.
+  #
+  # Default is `env`.
+  #
+  enabled: env
```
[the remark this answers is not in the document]
reivilibre: Sure, I suppose I did it because it feels wrong to me to have magic configuration that I can't keep track of, but I suppose anyone with a proxy will prefer to have the proxy declared in one place and everything else 'just work'™.

However, I think it'd be sensible to issue a warning if both are set (or maybe both are set differently?)? Does this sound good?

sygnal.yaml.sample
```diff
@@ -123,6 +123,31 @@ http:
...
+
+  # Specify an address and port to use for connecting to the HTTP proxy.
+  #
+  address: '127.0.0.1:8080'
+
+  # If your proxy uses HTTP proxy basic authentication, specify the username
+  # and password here.
+  #
+  #username:
+  #password:
```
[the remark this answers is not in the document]
reivilibre: Wilco

sygnal.yaml.sample
```diff
@@ -123,6 +123,31 @@ http:
...
+  #   `HTTPS_PROXY`/`https_proxy` environment variables.
+  # - `false` to be strict about not using one, not even from the environment.
+  #
+  # Default is `env`.
+  #
+  enabled: env
+
+  # Specify an address and port to use for connecting to the HTTP proxy.
+  #
+  address: '127.0.0.1:8080'
```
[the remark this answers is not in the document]
reivilibre: Yeah, I suppose this was less clear than I imagined.

sygnal/apnspushkin.py
reivilibre: Whoops, I didn't realise it was private! I'll change it.

sygnal/gcmpushkin.py
reivilibre: That is a host:port but it'll be going away

sygnal/gcmpushkin.py
reivilibre: Yes, it is, wonder how that came about, again… (memory hazy but I think it showed up during testing with proxies).

I'll take this out and rePR it if it's worthwhile.

reivilibre:
> Can we not have one that can be used in both places (not sure if it's easier to use the asyncio one or the twisted one)?

I've looked into this a lot, and my general feeling is that it would be making a trade;

the complexity of having 2 implementations is quite small (in my opinion); trying to create some Asyncio-to-Twisted Protocol wrapper seemed messy (and probably would lead to more code if anything).

I say this coming from not much Twisted experience, but the way there are 5 classes to do it in Twisted code is making me think this is complex enough and probably not clearly remappable with asyncio.

sygnal/helper/proxy/proxy_asyncio.py
```diff
@@ -0,0 +1,241 @@
...
+                    self.transport.close()
+                    raise ProxyConnectError(
+                        "Error from HTTP Proxy"
+                        f" whilst attempting CONNECT: {status.decode()}"
+                        f" ({reason_phrase.decode()}); aborting connection."
+                    )
+
+                logger.debug("Ready to switch over protocol")
+
+                self.buffer = None  # type: ignore
```
richvdh, line 124: why do we do this, out of interest?

Also: what happens to any bytes in the buffer after the CONNECT response? don't they need to go to the new protocol?

sygnal/helper/proxy/proxy_asyncio.py
```diff
@@ -0,0 +1,241 @@
...
+        self.proxy_url_parts = proxy_url_parts
+
+        # This future is completed when it is safe to take back control of the
+        # transport (which is also returned to indicate this).
+        self.wait_for_establishment: Future[Transport] = Future()
+
+    def data_received(self, data: bytes) -> None:
+        super().data_received(data)
+        self.buffer += data
+        if b"\r\n\r\n" in self.buffer:
```
richvdh, line 72: as a general rule: I'm a fan of early return in this sort of case. It leads to less indentation, and, more importantly, less mental load (I don't need to scroll all the way down to find out what happens in the `else` case; when I'm reading the rest of this function, I know that we're just handling the case that we've read all the headers).

Alternatively, factor out the whole long `if` body into a separate function, but early return is generally easier.

```suggestion
        if b"\r\n\r\n" not in self.buffer:
            # haven't finished the headers yet
            return
```
```

File: sygnal/helper/proxy/proxy_asyncio.py [hunk 17]
```diff
@@ -0,0 +1,241 @@
...
+        # (N.B. if we want to ever use TLS to the proxy [e.g. to protect the proxy
+        # credentials], we can ask this to give us a TLS connection).
+        transport, connect_protocol = await self._wrapped_loop.create_connection(
+            make_protocol, proxy_url_parts.hostname, proxy_url_parts.port
+        )
+
+        assert isinstance(connect_protocol, HttpConnectProtocol)
+
+        # wait for the HTTP Proxy CONNECT sequence to complete
+        transport = await connect_protocol.wait_for_establishment
```
richvdh, line 212: is there any reason to imagine that this `transport` could be different to the previous `transport`? if so, could you add a comment explaining? If not, I'd suggest :

```suggestion
        await connect_protocol.wait_for_establishment
```

sygnal/helper/proxy/proxy_asyncio.py
```diff
@@ -0,0 +1,241 @@
...
+            transport = await self._wrapped_loop.start_tls(
+                transport, connect_protocol, sslcontext
+            )
+
+        # construct the desired protocol and hand over the transport to it
+        new_protocol = protocol_factory()
+        # wire up transport to call `data_received` etc. on the new transport
+        transport.set_protocol(new_protocol)
+        # let the protocol know it has been connected to the transport
+        new_protocol.connection_made(transport)
```
richvdh, line 232: slightly wondering if it would be better to pass `sslcontext` and `protocol_factory` into HttpConnectProtocol's constructor, and moving this logic inside `HttpConnectProtocol.data_received`. It might make handling dangling data in the `buffer` easier, and would give you a CONNECT implementation that can be used without having to wrap the event loop, if that is ever useful.

sygnal/helper/proxy/proxy_asyncio.py
```diff
@@ -0,0 +1,241 @@
+import asyncio
```
richvdh, line 1: This could do with a copyright notice. It's hazy as to whether it should be matrix foundation or NVL, but I think you're doing this with a matrix.org hat on so matrix foundation.

richvdh: (likewise other new files that don't have a notice)

sygnal.yaml.sample
```diff
@@ -123,6 +123,20 @@ http:
...
+  # Specify the URL of a proxy to use for outgoing traffic
+  # (e.g. to Apple & Google) if desired.
+  # Currently only HTTP proxies with CONNECT capability are supported.
+  #
+  # If you do not specify a value, the `HTTPS_PROXY` environment variable will
+  # be used if present. Otherwise, no proxy will be used.
+  #
+  # Default is unspecified.
+  #
+  #url: 'http://user:secret@prox:8080'
```
richvdh, line 138: I'm still a bit baffled by this being a dict. Why not just 

```
# proxy: 'http://user:secret@prox:8080'
```
?

sygnal/apnspushkin.py
```diff
@@ -131,8 +132,18 @@ def __init__(self, name, sygnal, config):
             if not self.get_config("topic"):
                 raise PushkinSetupException("You must supply topic.")
 
+        # use the Sygnal global proxy configuration
+        proxy_url_str = sygnal.config["proxy"].get("url")
```
richvdh, line 136: won't this fail on existing configs with no `proxy` section?

sygnal/exceptions.py
```diff
@@ -34,3 +35,7 @@ class TemporaryNotificationDispatchException(Exception):
     def __init__(self, *args: object, custom_retry_delay=None) -> None:
         super().__init__(*args)
         self.custom_retry_delay = custom_retry_delay
+
+
+class ProxyConnectError(ConnectError):
```
richvdh, line 40: a docstring describing what this means would be good.

sygnal/helper/proxy/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+
+def decompose_http_proxy_url(proxy_url):
+    url = urlparse(proxy_url, scheme="http")
+
+    if not url.hostname:
+        raise RuntimeError("Proxy URL did not contain a hostname! Please specify one.")
+
+    if url.scheme != "http":
+        raise RuntimeError(
+            f"Unknown proxy scheme {url.scheme};" f" only 'http' is supported."
```
richvdh, line 12: I'm not familiar with `f""` syntax; how is this different from

```suggestion
            f"Unknown proxy scheme {url.scheme}; only 'http' is supported."
```

?

sygnal/helper/proxy/connectproxyclient_twisted.py
```diff
@@ -0,0 +1,215 @@
...
+        # CONNECT to complete.
+        d.addCallback(lambda conn: f.on_connection)
+        return d
+
+
+class HTTPProxiedClientFactory(protocol.ClientFactory):
+    """ClientFactory wrapper that triggers an HTTP proxy CONNECT on connect.
+
+    Once the CONNECT completes, invokes the original ClientFactory to build the
+    HTTP Protocol object and run the rest of the connection.
```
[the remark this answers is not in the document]
richvdh: hrm, cargoculted from synapse. What idiot wrote that? :innocent: 

could you fix it up here?

sygnal/helper/proxy/__init__.py
```diff
@@ -0,0 +1,15 @@
+from urllib.parse import urlparse
+
+
+def decompose_http_proxy_url(proxy_url):
```
richvdh, line 4: could do with a docstring and some type annotations. bytes, str or either?

sygnal/helper/proxy/connectproxyclient_twisted.py
```diff
@@ -0,0 +1,215 @@
...
+
+    Args:
+        dst_host (bytes): hostname that we want to CONNECT to
+        dst_port (int): port that we want to connect to
+        wrapped_factory (protocol.ClientFactory): The original Factory
+    """
+
+    def __init__(self, dst_host, dst_port, proxy_url_parts, wrapped_factory):
```
richvdh, line 86: proxy_url_parts is generally underdocumented in this whole file.

It feels a bit odd that we're passing the whole of `proxy_url_parts` around, when most of it is redundant. Why not just an optional `Tuple[str, str]` for the username/password?

sygnal/helper/proxy/connectproxyclient_twisted.py
```diff
@@ -0,0 +1,215 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 The Matrix.org Foundation C.I.C.
```
richvdh, line 2: You've made additions here, so:

```suggestion
# Copyright 2019-2020 The Matrix.org Foundation C.I.C.
```

sygnal/helper/proxy/proxyagent_twisted.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+
+    def __init__(
+        self,
+        reactor,
+        contextFactory=BrowserLikePolicyForHTTPS(),
+        connectTimeout=None,
+        bindAddress=None,
+        pool=None,
+        proxy_url_str: str = None,
```
richvdh, line 69: a str can't be none

```suggestion
        proxy_url_str: Optional[str] = None,
```

richvdh: please can haz docstring for `proxy_url_str` ?

sygnal/helper/proxy/proxyagent_twisted.py
```diff
@@ -0,0 +1,183 @@
...
+
+        self._endpoint_kwargs = {}
+        if connectTimeout is not None:
+            self._endpoint_kwargs["timeout"] = connectTimeout
+        if bindAddress is not None:
+            self._endpoint_kwargs["bindAddress"] = bindAddress
+
+        self.proxy_endpoint = _http_proxy_endpoint(
+            proxy_url_str, reactor, **self._endpoint_kwargs
+        )
```
richvdh, line 81: `_http_proxy_endpoint` is only three lines, of which two are here already. Wouldn't it be easier to inline it?

README.rst
```diff
@@ -65,8 +65,8 @@ gcm
   to contain the 'Server key', which can be acquired from Firebase Console at:
   ``https://console.firebase.google.com/project/<PROJECT NAME>/settings/cloudmessaging/``
 
-Pusher `data` configuration
-=======
+Pusher ``data`` configuration
+-----------------------------
```
richvdh, line 69: unlike your new section, this section doesn't belong under `Setup`. You need to undo this and move the new section up.

richvdh:
> the complexity of having 2 implementations is quite small (in my opinion, even if it is a bit sad); trying to create some Asyncio-to-Twisted Protocol wrapper seemed messy (and probably would lead to more code if anything).

I feel like the easiest way might be to make some sort of Transport adaptor so that you can connect the twisted HTTP11ClientProtocol to the asyncio TCPTransport. But I'm inclined to agree that it is likely to be fiddly, and TLS is likely to make everything an order of magnitude harder.

Thanks for looking into it, anyway.

sygnal/helper/proxy/proxy_asyncio.py
reivilibre: Very true, wish I spotted these things easier :)

README.rst
reivilibre: Suppose I should have checked / left as-is but felt very 'Setup'-y to me.

sygnal/apnspushkin.py
reivilibre: No, the defaults take care of that

sygnal/helper/proxy/__init__.py
reivilibre: It isn't (nothing to do with the f-ness in this case), unfortunately it often happens that I/my IDE splits multiple lines into e.g.

```python
spqr(
    "This is a long line of text addressed to"
    " Monsieur vdH"
)
```

then if e.g. indentation changes later on, `black` will move them to the same line but not rejoin them together (and I often don't spot them).

sygnal/helper/proxy/proxy_asyncio.py
reivilibre: > why do we do this, out of interest?

Probably not concerning in the grand scheme of things, but intended to allow the GC to take back the memory spent by the byte buffer.

Would it be less silly looking to use `b""` instead of None?

> Also: what happens to any bytes in the buffer after the CONNECT response? don't they need to go to the new protocol?

In a totally generic world, yes.

But neither HTTP nor TLS allows the server to speak before the client (this is mentioned a little bit up, the context shifted downwards throughout tweaking, I suppose), so the situation is that there are no bytes left over (should I assert this or figure out a way to pass back the theoretical leftovers?).

sygnal/helper/proxy/proxy_asyncio.py
reivilibre: Yes, I agree, but `loop.start_tls(` is an `async def`, whereas none of the stuff in `Protocol` is `async`, it's all callback driven.

reivilibre: I think I have a plan

sygnal/helper/proxy/proxy_asyncio.py
richvdh: > But neither HTTP nor TLS allows the server to speak before the client

are you sure about this? when I `openssl s_client -connect api.push.apple.com:443`, I get some bytes spat back at me.

reivilibre: > are you sure about this? when I `openssl s_client -connect api.push.apple.com:443`, I get some bytes spat back at me.

Me too. I suppose I need to update my assumptions since HTTP/2 … it seems like quite a large rework of HTTP.

I would say I expect TLS will still be fine (and `nc api.push.apple.com 443` doesn't show anything come back before I start typing) but it may just be a matter of time until someone breaks that assumption.

Lucky my new version can handle dangling bytes then … :)

sygnal/helper/proxy/proxy_asyncio.py
```diff
@@ -0,0 +1,321 @@
...
+            raise ValueError("transport must be a proper Transport")
+
+        super().connection_made(transport)
+        # when we get a TCP connection to the HTTP proxy, we invoke the CONNECT
+        # method on it to open a tunnelled TCP connection through the proxy to
+        # the other side
+        host, port = self.target_hostport
+        transport.write(f"CONNECT {host}:{port} HTTP/1.1\r\n".encode())
+        parts = self.proxy_url_parts
+        transport.write(f"Host: {parts.hostname}:{parts.port or 80}\r\n".encode())
```
richvdh, line 224: wrong default port. Also, I'm pretty sure that convention is that if it's the default port (ie, 443 or 80, depending on your protocol) then you omit the port from the Host header.

Also... it looks like the twisted impl doesn't bother with this? if it's not needed that would simplify the interface.

sygnal/helper/proxy/proxy_asyncio.py
```diff
@@ -0,0 +1,321 @@
...
+
+        # enforced by decompose_http_proxy_url
+        assert proxy_url_parts.hostname is not None
+
+        # create a raw TCP connection to the proxy
+        # (N.B. if we want to ever use TLS to the proxy [e.g. to protect the proxy
+        # credentials], we can ask this to give us a TLS connection).
+
+        transport, connect_protocol = await self._wrapped_loop.create_connection(
+            make_protocol, proxy_url_parts.hostname, proxy_url_parts.port or 80
```
richvdh, line 305: wrong default port

sygnal/helper/proxy/proxyagent_twisted.py
```diff
@@ -0,0 +1,170 @@
...
+            parsed_url = decompose_http_proxy_url(proxy_url_str)
+            if parsed_url.username is None or parsed_url.password is None:
+                self._proxy_auth = None
+            else:
+                self._proxy_auth = (parsed_url.username, parsed_url.password)
+
+            self.proxy_endpoint = HostnameEndpoint(
+                reactor,
+                parsed_url.hostname,
+                parsed_url.port or 80,
```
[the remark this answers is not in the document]
reivilibre: I'm not sure I can agree with that; 1080 is [registered for SOCKS proxies](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=1080)

> Also, shouldn't this be sorted out in decompose_http_proxy_url ?

Good point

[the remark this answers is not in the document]
reivilibre: (Firefox uses port 80 if not specified, as a data point)

sygnal/helper/proxy/proxy_asyncio.py
reivilibre: There is no requirement to omit the port if it is default. It may be cleaner but not necessary; that said, should I bother anyway?

However, in my experience netcatting HTTP servers, they choke (400?) if you leave off the Host header... [This is backed up by MDN.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host)

sygnal/helper/proxy/proxyagent_twisted.py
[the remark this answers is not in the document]
richvdh: hrm fair enough. certainly when it's spelt with an `http://` scheme, port 80 seems more sensible.

either way, sticking it in `decompose_http_proxy_url` will stop it being repeated 3 times.

sygnal/helper/proxy/proxy_asyncio.py
richvdh: http servers yes; for CONNECT proxies I think the case is a bit different.

anyway if you want to leave it as is, that's fine.

richvdh: Interesting that twisted.web makes special effort to skip the default port: https://github.com/twisted/twisted/blob/d47dafe0717b3877d5d0017ab0feb0721cb78613/src/twisted/web/http.py#L1468-L1476

I kinda think it would be best to go with the flow here and omit it for the default port.

sygnal/helper/proxy/proxy_asyncio.py
```diff
@@ -0,0 +1,321 @@
...
+            if status != b"200":
+                # 200 Successful (aka Connection Established) is what we want
+                # if it is not what we have, then we don't have a tunnel
+                logger.error(
+                    "Error from HTTP Proxy"
+                    " whilst attempting CONNECT: %s (%s);"
+                    "aborting connection.",
+                    status,
+                    reason_phrase,
+                )
```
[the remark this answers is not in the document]
reivilibre: have dealt with, would appreciate it if you checked this was satisfactory

sygnal/helper/proxy/proxy_asyncio.py
reivilibre: I notice that Firefox will set the `Host:` to the same as the `CONNECT` destination.
I can't find any results suggesting that Host is optional for HTTP proxies...
The code nicked from Synapse uses HTTP/1.0 and Host isn't required for HTTP/1.0.

So I'm inclined to go with the flow and switch over to HTTP/1.0 and take away the Host header, in case it has a different meaning to what I expected and causes someone's proxy to get upset.

reivilibre:
Going to work on tests but happy to have a concurrent review if you have time.

PR #131, opened 2020-07-02
Make it MyPy and isort-friendly

reivilibre:
Needs a change in the matrix-org/pipelines to enable in CI too.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,6 +1,6 @@
 [mypy]
 namespace_packages = True
-plugins = mypy_zope:plugin
+# plugins = mypy_zope:plugin
 follow_imports = silent
 check_untyped_defs = True
 show_error_codes = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,7 +1,4 @@
 [mypy]
-namespace_packages = True
-# plugins = mypy_zope:plugin
-follow_imports = silent
 check_untyped_defs = True
 show_error_codes = True
 show_traceback = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,35 @@
+[mypy]
+namespace_packages = True
+plugins = mypy_zope:plugin
+follow_imports = silent
+check_untyped_defs = True
+show_error_codes = True
+show_traceback = True
+mypy_path = stubs
+
+[mypy-idna]
+ignore_missing_imports = True
+
+[mypy-OpenSSL]
+ignore_missing_imports = True
+
+[mypy-opentracing.*]
+ignore_missing_imports = True
+
+[mypy-twisted.*]
+ignore_missing_imports = True
+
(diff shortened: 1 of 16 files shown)


sygnal/apnstruncate.py
```diff
@@ -85,7 +86,7 @@ def truncate(payload, max_length=2048):
 
 
 def _choppables_for_aps(aps):
-    ret = []
+    ret: List[Union[Tuple[str], Tuple[str, int]]] = []
```
clokep, line 89: I think we've usually be using the `# type: Foo` style in Synapse, I believe this is due to Python 3.5 not supporting this syntax? I don't know if that's supported for Syngal.

Also 👍 on adding this!

reivilibre: Since the start of the Sygnal rewrite, the readme has said we require Python 3.7+, so should be good :)

sygnal/notifications.py
```diff
@@ -63,24 +64,18 @@ def __init__(self, raw):
...
-                self.__dict__[a] = notif[a]
-            else:
-                self.__dict__[a] = None
+        # optional attributes
+        self.room_name: Optional[str] = notif.get("room_name")
+        self.room_alias: Optional[str] = notif.get("room_alias")
+        self.prio: Optional[str] = notif.get("prio")
+        self.membership: Optional[str] = notif.get("membership")
+        self.sender_display_name: Optional[str] = notif.get("sender_display_name")
+        self.content = notif.get("content")
```
clokep, line 73: I think you missed this one?

```suggestion
        self.content: Optional[str] = notif.get("content")
```

reivilibre: I did miss that one, but the type of content appears to be a dict (I should check the spec to be sure that there is no other option on top of that, though).

mypy.ini
```diff
@@ -0,0 +1,35 @@
+[mypy]
+namespace_packages = True
+# plugins = mypy_zope:plugin
```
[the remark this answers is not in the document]
reivilibre: It was, I should probably read through some of these things in the `[mypy]` section and trim down...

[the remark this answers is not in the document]
reivilibre: I had a look and it was probably worthwhile as Synapse had some errors silenced (probably because it wasn't feasible to solve them at that scale).

sygnal/apnspushkin.py
```diff
@@ -220,7 +221,7 @@ async def dispatch_notification(self, n, device, context):
         # The pushkey is kind of secret because you can use it to send push
         # to someone.
         # span_tags = {"pushkey": device.pushkey}
-        span_tags = {}
+        span_tags: Dict[str, Union[str, int]] = {}
```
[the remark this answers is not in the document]
reivilibre: You're right afaict

Issue #117, opened 2020-05-11
apns requests seem to stack up forever on connection failure


richvdh:
When our TLS certificate expired, so that all apns requests were unsuccessful, the "requests in flight" and "apns requests in flight" metrics went up seemingly without limit:

![image](https://user-images.githubusercontent.com/1389908/81584723-0624ff00-93ab-11ea-8f15-a87cfe0acd52.png)

I would expect that we retry each request a few (well, 12) times, and then give up (see also #116), so I'd expect requests to stack up a bit, but not to this extent.

We ended up with a 5G RSZ, so I don't think this is just a problem with the metrics.

richvdh:
see https://github.com/matrix-org/sygnal/issues/104 for a bit more background here.

PR #132, opened 2020-07-08
In-flight request limiting

reivilibre:
Intended to provide a general increase in robustness/fault-tolerance, e.g. to address #117 .


reivilibre:
Thanks, I'll bear it in mind :)

PR #133, opened 2020-07-08
Don't try so hard, APNS — 3 times only, like GCM

### LINKED ISSUE #116: consider doing fewer retries for APNS
if we have problems with APNS, we retry each request at 10s and 20s intervals. Given that haproxy limits the number of concurrent requests to sygnal, this can easily mean that the haproxy<->signal connection pool gets saturated with failing APNS requests, so that very few GCM requests get a look-in.

We should consider doing fewer retries, or doing them faster, or marking APNS as down and failing the request more quickly.

Really, we need a mechanism to indicate to homeservers that a push failed (https://github.com/matrix-org/matrix-doc/issues/2534).

reivilibre:
Signed-off-by: Olivier Wilkinson (reivilibre) <olivier@librepush.net>

This makes a minor difference so that APNS has the same retry schedule as GCM (without extra retries-per-retry).

It may still be doing too much (hopefully mitigated by #132 though)

Fixes #116

### DIFF
--- a/changelog.d/133.misc
+++ b/changelog.d/133.misc
@@ -1,0 +1,1 @@
+Don't try an aioapns connection 4 times per APNS dispatch attempt; do the same as GCM.
--- a/changelog.d/133.misc
+++ b/changelog.d/133.misc
@@ -1,1 +1,1 @@
-Don't try an aioapns connection 4 times per APNS dispatch attempt; do the same as GCM.
+Attempt the same number of retries for both GCM and APNS.
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -132,6 +132,9 @@ class ApnsPushkin(Pushkin):
                 raise PushkinSetupException("You must supply topic.")
 
         if certfile is not None:
+            # max_connection_attempts is actually the maximum number of
+            # additional connection attempts, so =0 means try once only
+            # (we will retry at a higher level so not worth doing more here)
             self.apns_client = APNs(
                 client_cert=certfile,
                 use_sandbox=self.use_sandbox,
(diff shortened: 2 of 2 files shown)


changelog.d/133.misc
```diff
@@ -0,0 +1 @@
+Don't try an aioapns connection 4 times per APNS dispatch attempt; do the same as GCM.
```
clokep, line 1: Maybe:

> Attempt the same number of retries for both GCM and APNS.

sygnal/apnspushkin.py
```diff
@@ -132,7 +132,11 @@ def __init__(self, name, sygnal, config):
                 raise PushkinSetupException("You must supply topic.")
 
         if certfile is not None:
-            self.apns_client = APNs(client_cert=certfile, use_sandbox=self.use_sandbox)
+            self.apns_client = APNs(
+                client_cert=certfile,
+                use_sandbox=self.use_sandbox,
+                max_connection_attempts=0,
```
clokep, line 138: I'm guessing that this parameter really means "max *additional* connection attempts if the first fails"?

This might need a comment saying that there's a number of connection attempts, but also a number of retries that occur?

changelog.d/133.misc
reivilibre: How dare you suggest that we write something non-cryptic in a changelog? :)

I agree, will change this over, I must have had some kind of commit writer's block at the time :S!

sygnal/apnspushkin.py
reivilibre: Yes, could do with a comment :)

changelog.d/133.misc
clokep: It's hard to write it from the POV of someone running sydent instead of what the internal changes mean!

reivilibre:
Thanks, this is back in review :)

clokep:
@reivilibre Looks like CI is being funky here?

reivilibre:
Upsy daisies, yes, this repo needed master pulling in for CI changes. :) Sorry!

PR #134, opened 2020-07-15
Support Tox for unit tests and linting

reivilibre:
We will then use this in CI.

I also updated the CONTRIBUTING documentation to reflect this, and fixed tests so they pass mypy (which is both good for us and makes it easier to use the `lint.sh` script).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py37, check_isort, check_codestyle, check_types
+envlist = py37, check_codestyle, check_types
 
 [testenv]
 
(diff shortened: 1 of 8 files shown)


tox.ini
```diff
@@ -0,0 +1,34 @@
...
+# creates a symlink to the project directory instead of unpacking the sdist.
+usedevelop=true
+
+commands = trial tests
+
+[testenv:check_codestyle]
+skip_install = True
+deps =
+  # pin deps so we don't start failing when they are updated
+  flake8==3.8.3
```
clokep, line 21: Not sure if it matters, but we also use `flake8-comprehensions` on Synapse. Probably should be introduced as a separate PR though.

PR #139, opened 2020-07-27
Update to aioapns 1.10 for security fixes

reivilibre:
Signed-off-by: Olivier Wilkinson (reivilibre) <olivier@librepush.net>


changelog.d/139.bugfix
```diff
@@ -0,0 +1 @@
+Update to aioapns 1.10 which has security fixes.
```
richvdh, line 1: this is a bit misleading. there was nothing stopping us using 1.10 before.

```suggestion
Update minimum version of `aioapns` dependency to 1.10, which has security fixes.
```

PR #145, opened 2020-08-05
Add configuration options to allow specifying FCM options

reivilibre:
This is practically a must-have for anyone who wants to use Firebase with iOS, as content-available is needed to cause an iOS app to wake up from an APNs push.

### DIFF
--- a/changelog.d/145.feature
+++ b/changelog.d/145.feature
@@ -1,0 +1,1 @@
+Add ability to configure custom FCM options, which can be useful for using iOS with Firebase.
--- a/changelog.d/145.feature
+++ b/changelog.d/145.feature
@@ -1,1 +1,1 @@
-Add ability to configure custom FCM options, which can be useful for using iOS with Firebase.
+Add the ability to configure custom FCM options, which is necessary for using iOS with Firebase.
--- a/sygnal.yaml.sample
+++ b/sygnal.yaml.sample
@@ -252,8 +252,16 @@ apps:
   #
   #  # This allows you to specify additional options to send to Firebase.
   #  # See https://firebase.google.com/docs/cloud-messaging/http-server-ref
+  #  # for the exhaustive list of valid options.
   #  # Of particular interest, admins who wish to support iOS apps using Firebase
   #  # probably wish to set content_available, and may need to set mutable_content.
+  #  # (content_available allows your iOS app to be woken up by data messages)
+  #  # and mutable_content allows your notification to be modified by a
+  #  # Notification Service app extension).
+  #  #
+  #  # Do not specify `data`, `priority`, `to` or `registration_ids` as they may
+  #  # be overwritten or lead to an invalid request.
+  #  #
   #  #fcm_options:
   #  #  content_available: true
   #  #  mutable_content: true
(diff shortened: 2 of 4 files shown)


sygnal.yaml.sample
```diff
@@ -249,3 +249,11 @@ apps:
   #  # Defaults to 512 if unset.
   #  #
   #  #inflight_request_limit: 512
+  #
+  #  # This allows you to specify additional options to send to Firebase.
+  #  # See https://firebase.google.com/docs/cloud-messaging/http-server-ref
+  #  # Of particular interest, admins who wish to support iOS apps using Firebase
+  #  # probably wish to set content_available, and may need to set mutable_content.
```
anoadragon453, line 256: Is it absolutely required? Why would I do this for my (theoretical) setup?

sygnal/gcmpushkin.py
```diff
@@ -134,6 +135,12 @@ def __init__(self, name, sygnal, config, canonical_reg_id_store):
         if not self.api_key:
             raise PushkinSetupException("No API key set in config")
 
+        self.base_request_body: dict = self.get_config("fcm_options", {})
+        if not isinstance(self.base_request_body, dict):
+            raise PushkinSetupException(
+                "fcm_options, if set, should be a dictionary of options."
```
anoadragon453, line 141:
```suggestion
                "Config field fcm_options, if set, must be a dictionary of options"
```

sygnal/gcmpushkin.py
```diff
@@ -348,7 +355,11 @@ async def _dispatch_notification_unlimited(self, n, device, context):
             # TODO: Implement collapse_key to queue only one message per room.
             failed = []
 
-            body = {"data": data, "priority": "normal" if n.prio == "low" else "high"}
+            # use the base body as a foundation for the body; this lets the
+            # Sygnal admin choose custom FCM options (e.g. content_available).
```
anoadragon453, line 359: I think this comment might be better suited to where we pull the options from the config. Perhaps with some rewording of `Use the fcm_options config dictionary as a foundation for the body...`

sygnal.yaml.sample
```diff
@@ -249,3 +249,11 @@ apps:
   #  # Defaults to 512 if unset.
   #  #
   #  #inflight_request_limit: 512
+  #
+  #  # This allows you to specify additional options to send to Firebase.
```
anoadragon453, line 253: You may want to be explicit that admins can specify *any* option from the linked URL here, not just those that are listed below (which is what I initially thought).

sygnal.yaml.sample
```diff
@@ -249,3 +249,11 @@ apps:
...
+  #  # See https://firebase.google.com/docs/cloud-messaging/http-server-ref
+  #  # Of particular interest, admins who wish to support iOS apps using Firebase
+  #  # probably wish to set content_available, and may need to set mutable_content.
+  #  #fcm_options:
+  #  #  content_available: true
+  #  #  mutable_content: true
```
anoadragon453, line 259: It's not clear that if `priority` or `data` (for whatever reason it'd be specified here) is set, its value will be overwritten later.

changelog.d/145.feature
```diff
@@ -0,0 +1 @@
+Add ability to configure custom FCM options, which can be useful for using iOS with Firebase.
```
anoadragon453, line 1:
```suggestion
Add the ability to configure custom FCM options, which is necessary for using iOS with Firebase.
```

?

reivilibre: yeah, fair enough

PR #146, opened 2020-08-06
Add Prometheus metric for sygnal_inflight_request_limit_drop

reivilibre:


sygnal/notifications.py
```diff
@@ -142,6 +144,13 @@ class ConcurrencyLimitedPushkin(Pushkin):
 
     UNDERSTOOD_CONFIG_FIELDS = {"inflight_request_limit"}
 
+    RATELIMITING_DROPPED_REQUESTS = Counter(
```
richvdh, line 147: question: if there are two pushkins (say, for GCM and APNS), do they both correctly refer to the same instance of `Counter` ? or do they each get their own instance (which will probably confuse the prometheus lib)?

(I can never quite remember how class-level attributes work)

clokep: They should get the same one. I wasn't 100% sure that's what was wanted, but I *think* it is.

reivilibre: My understanding was/is that the class attribute is initialised when the class declaration is executed, and that attribute stays with the class.

Note it is always referred to as `ConcurrencyLimitedPushkin.RATELIMITING_DROPPED_REQUESTS`, so it would always act like a namespaced global.

I didn't know whether subclasses would get access to it (not that it matters, since we access the field statically every time and not through `self` or `cls`), so I tried it:

```python
oli@bbm-neon:~$ python3
Python 3.6.9 (default, Jul 17 2020, 12:50:27) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> class X:
...     bob = []
... 
>>> class Y(X):
...     pass
... 
>>> class Z(X):
...     pass
... 
>>> id(Y.bob)
140481390301896
>>> id(X.bob)
140481390301896
>>> id(Z.bob)
140481390301896
```

reivilibre: (and to answer the question, yes this is intended — the pushkin is distinguished by the `pushkin` label, as this is what Prometheus understands)

clokep: In your example, even doing the following should return the same thing:

```python
id(Y().bob)
```

Doesn't matter if you access it via the class name or not (although I think it is clearer to access class-level variables that way).

reivilibre: > In your example, even doing the following should return the same thing:
> 
> ```python
> id(Y().bob)
> ```
> 
> Doesn't matter if you access it via the class name or not (although I think it is clearer to access class-level variables that way).

Yes you are right (and I suspected so), but my experience from Java is that 'access static vars by class name, even if you can do it by instance' :)

PR #151, opened 2020-08-25
Tests for the HTTP Proxy support (asyncio and common only)

reivilibre:
This adds:
- tests for the HTTP proxy support for pushkins that use asyncio
- tests for the common code as part of HTTP proxy support — namely, the proxy URL parsing logic

#152 contains the Twisted stuff, majority of which is lifted from Synapse


clokep:
@reivilibre Can you add a description to this (and how it differs from #152? Is it built on top of it? Is it separate?)

reivilibre:
@clokep Thanks for looking. It's fully independent (owing to the fact we sadly have two independent proxy client implementations in here because 'Twisted is a thing but so is asyncio'). Added comments above ^.

This one should be less grubby than the Twisted stuff (though this code here is original, the other is blame Synapse :)).

Hope that helps.

tests/test_proxy_url_parsing.py
```diff
@@ -0,0 +1,45 @@
...
+        self.assertEqual(
+            parts, HttpProxyUrl("example.org", 80, ("bob", "secretsquirrel"))
+        )
+
+        parts = decompose_http_proxy_url("http://bob:secretsquirrel@example.org:8080")
+        self.assertEqual(
+            parts, HttpProxyUrl("example.org", 8080, ("bob", "secretsquirrel"))
+        )
+
+    def test_decompose_http_proxy_url_failure(self):
```
clokep, line 38: This could use a test for providing just a username? Or maybe that's not an error, but no credentials are set-up?

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+from tests.asyncio_test_helpers import (
+    MockProtocol,
+    MockTransport,
+    TimelessEventLoopWrapper,
+)
+
+
+class AsyncioHttpProxyTest(testutils.TestCase):
+    def config_setup(self, config):
+        super(AsyncioHttpProxyTest, self).config_setup(config)
```
clokep, line 32:
```suggestion
        super().config_setup(config)
```

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+        config["apps"]["com.example.spqr"] = {
+            "type": "tests.test_pushgateway_api_v1.TestPushkin"
+        }
+        base_loop = asyncio.new_event_loop()
+        augmented_loop = TimelessEventLoopWrapper(base_loop)  # type: ignore
+        asyncio.set_event_loop(cast(AbstractEventLoop, augmented_loop))
```
clokep, line 38: I'm just assuming that this mostly works.

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+            protocol_factory=lambda: fake_protocol,
+            sslcontext=None,
+            loop=None,
+        )
+
+        switch_over_task = asyncio.get_event_loop().create_task(
+            hcp.switch_over_when_ready()
+        )
+
+        # if it's already done, that'd be odd
```
clokep, line 64: Odd...or an error? Maybe making this comment stronger?

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+        """
+        Tests the proxy connection procedure when there is no basic auth.
+        """
+        # make a fake proxy
+        fake_proxy = MockTransport()
+        # make a fake protocol that we fancy using through the proxy
+        fake_protocol = MockProtocol()
+
+        # create a connect proxy client
+        hcp = HttpConnectProtocol(
```
clokep, line 52: `connect_protocol` might be a better name than `hcp`.

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+        )
+
+        # if it's already done, that'd be odd
+        self.assertFalse(switch_over_task.done())
+
+        # connect the proxy client to the proxy
+        fake_proxy.set_protocol(hcp)
+        hcp.connection_made(fake_proxy)
+
+        # check we got sent a reasonable, uh, hardcoded, CONNECT request
```
clokep, line 71:
```suggestion
        # Check that the proxy got the proper CONNECT request.
```

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+        self.assertEqual(fake_proxy.buffer, b"CONNECT example.org:443 HTTP/1.0\r\n\r\n")
+        # clear
```
clokep, line 73:
```suggestion
        # Reset the proxy.
```

I usually like to make a `reset_mock()` method on fake objects that handle things like this. It [matches `Mock` from the standard library](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.reset_mock).

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+        fake_proxy.buffer = b""
+
+        # pretend we got a happy response with some dangling bytes from the
+        # target protocol
+        fake_proxy.pretend_to_receive(
+            b"HTTP/1.0 200 Connection Established\r\n\r\n"
+            b"begin beep boop\r\n\r\n~~ :) ~~"
+        )
+
+        # advance reactor because we have to let coroutines be executed
```
clokep, line 83: Is this actually a reactor?

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+        self.loop.advance(1.0)
+
+        # *now* we should have switched over!
+        self.assertTrue(switch_over_task.done())
```
clokep, line 87: I'm not sure what this means? What does "switched over" mean? Should this be "Now we should be using the proxy"?

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+
+        transport, protocol = switch_over_task.result()
+
+        # check it was our protocol that was returned
+        self.assertEqual(protocol, fake_protocol)
```
clokep, line 92: `assertIs` maybe?

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+        switch_over_task = asyncio.get_event_loop().create_task(
+            hcp.switch_over_when_ready()
+        )
+
+        # if it's already done, that'd be odd
+        self.assertFalse(switch_over_task.done())
+
+        # connect the proxy client to the proxy
+        fake_proxy.set_protocol(hcp)
+        hcp.connection_made(fake_proxy)
```
clokep, line 126: Looks like this can be in `setUp` or a helper method so we don't repeat the code.

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+
+        # check we got sent a reasonable, uh, hardcoded, CONNECT request
+        # with the correctly-encoded credentials
+        self.assertEqual(
+            fake_proxy.buffer,
+            b"CONNECT example.org:443 HTTP/1.0\r\n"
+            b"Proxy-Authorization: basic dXNlcjpzZWNyZXQ=\r\n\r\n",
```
clokep, line 133: Out of curiosity, how are these encoded?

tests/test_httpproxy_asyncio.py
```diff
@@ -0,0 +1,218 @@
...
+        )
```
clokep, line 197: What makes this invalid?

tests/asyncio_test_helpers.py
```diff
@@ -0,0 +1,162 @@
...
+from asyncio.transports import Transport
+from typing import Any, List, Tuple
+
+logger = logging.getLogger(__name__)
+
+
+class TimelessEventLoopWrapper:
+    @property  # type: ignore
+    # -- we have no need to make this read/write
+    # (o.O being able to re-write a __class__ is odd!)
```
clokep, line 14: I'm not following what this means. Please don't put comments between a decorator and a method -- it makes it hard to follow the code flow.

tests/asyncio_test_helpers.py
```diff
@@ -0,0 +1,162 @@
...
+    def __class__(self):
+        """
+        Fakes isinstance(this, AbstractEventLoop) so we can set_event_loop
+         without fail.
```
clokep, line 18:
```suggestion
        Fakes isinstance(this, AbstractEventLoop) so we can set_event_loop
        without fail.
```

tests/asyncio_test_helpers.py
```diff
@@ -0,0 +1,162 @@
...
+import types
+from asyncio import AbstractEventLoop, transports
+from asyncio.protocols import BaseProtocol, Protocol
+from asyncio.transports import Transport
+from typing import Any, List, Tuple
+
+logger = logging.getLogger(__name__)
+
+
+class TimelessEventLoopWrapper:
```
clokep, line 11: I'm surprised that Python doesn't come with a test asyncio loop?

tests/asyncio_test_helpers.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        We use this to delegate other method calls to the real EventLoop.
+        """
+        value = getattr(self._wrapped_loop, item)
+        if isinstance(value, types.MethodType):
+            # rebind this method to be called on us
+            # this makes the wrapped class use our overridden methods when
+            # available.
+            # Otherwise, call_soon etc won't come to us in all scenarios.
+            return types.MethodType(value.__func__, self)
```
clokep, line 59: This is because this isn't a sub-class of the "real" loop?

tests/asyncio_test_helpers.py
```diff
@@ -0,0 +1,162 @@
...
+        else:
+            return value
+
+    def call_later(self, delay, callback, *args, context=None):
+        self.call_at(self._time + delay, callback, *args)
```
clokep, line 64:
```suggestion
        self.call_at(self._time + delay, callback, *args, context)
```

tests/asyncio_test_helpers.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def call_at(self, when, callback, *args, context=None):
+        logger.debug(f"Calling {callback} at %f...", when)
+        self._to_be_called.append((when, callback, args, context))
+
+        # re-sort list in ascending time order
+        self._to_be_called.sort(key=lambda x: x[0])
```
clokep, line 71: I think sort will do what you want without `key` here since `when` is the first item in the tuple. 🤷

tests/asyncio_test_helpers.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def call_soon(self, callback, *args, context=None):
+        return self.call_later(0, callback, *args, context=context)
+
+    def time(self):
+        return self._time
+
+
+class MockTransport(Transport):
```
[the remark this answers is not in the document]
reivilibre: The Java programmer in me felt that this should be.
But you're right, the duck typingness of Python means we need not explicitly override Transport ourselves.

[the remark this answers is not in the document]
reivilibre: Oh:
```
builtins.ValueError: transport must be a proper Transport
```

I think Twisted devs also think in a Java way, a lot of the Twisted library gives off Java vibes anyway :). I don't think we can do anything clean about this, regardless of what is best.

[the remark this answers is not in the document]
reivilibre: Note that `Transport` is an interface (again, a Javaism).

I think it's somewhat clear that they'd consider changing the interface a breaking change in Twisted, much like it would in Java.

tests/asyncio_test_helpers.py
```diff
@@ -0,0 +1,162 @@
...
+    A transport intended to be driven by tests.
+    Stores received data into a buffer.
+    """
+
+    def __init__(self):
+        self.buffer = b""
+        self.eofed = False
+        self.aborted = False
+        self.protocol = None
+        self.closed = False
```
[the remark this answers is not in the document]
reivilibre: These are intended to be public to the code using this Mock in a test. They do not override anything (Transport seems much like a Java interface in that it has no fields. *Surely the Twisted devs have gone for a Java style... Flashbacks....*)

Would it be better if I put some comments on?

tests/test_httpproxy_asyncio.py
reivilibre: `base64(username:password)` :)

tests/asyncio_test_helpers.py
reivilibre: Me too :(

tests/asyncio_test_helpers.py
reivilibre: I would tend to agree except at the extreme:

If two events have the same time, then if we don't have key, it will sort by the callback.
I don't know if `call_later` is meant to be in-order — if it was, that would conflict with doing that.

tests/asyncio_test_helpers.py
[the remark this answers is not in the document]
clokep: Isn't this all asyncio, not Twisted?

tests/asyncio_test_helpers.py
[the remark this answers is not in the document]
clokep: I think that's OK, just making sure!

tests/test_httpproxy_asyncio.py
reivilibre: Oops, Twisted getting to me :)

tests/asyncio_test_helpers.py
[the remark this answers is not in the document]
reivilibre: …. urr, yes. I suppose the similarity has finally gotten to me and confused me. Oops.

tests/asyncio_test_helpers.py
clokep: Alright. 👍

tests/asyncio_test_helpers.py
reivilibre: Sorry, didn't know where to put the comment in that situation. Suspect it was mostly aimed at me in any case — took it out.

tests/asyncio_test_helpers.py
reivilibre: Yes, now clarified

tests/test_proxy_url_parsing.py
reivilibre: I added a test for username-only.

tests/asyncio_test_helpers.py
```diff
@@ -0,0 +1,176 @@
...
+            # event loop, which will call `self.call_later` or something
+            # which won't normally hit us because we are not an actual subtype.
+            return types.MethodType(value.__func__, self)
+        else:
+            return value
+
+    def call_later(self, delay: float, callback: Callable, *args: Any, context=None):
+        self.call_at(self._time + delay, callback, *args, context=context)
+
+    def call_at(self, when: float, callback: Callable, *args: Any, context=None):
```
clokep, line 66: Can we add a type hint for `context` to these few methods?

tests/asyncio_test_helpers.py
```diff
@@ -0,0 +1,176 @@
...
+        logger.debug(f"Calling {callback} at %f...", when)
+        self._to_be_called.append((when, callback, args, context))
+
+        # re-sort list in ascending time order
+        self._to_be_called.sort(key=lambda x: x[0])
+
+    def call_soon(self, callback: Callable, *args: Any, context=None):
+        return self.call_later(0, callback, *args, context=context)
+
+    def time(self):
```
clokep, line 76: Since the rest of the class has type hints. 😄 

```suggestion
    def time(self) -> float:
```

tests/asyncio_test_helpers.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class TimelessEventLoopWrapper:
+    @property  # type: ignore
+    def __class__(self):
+        """
+        Fakes isinstance(this, AbstractEventLoop) so we can set_event_loop
+        without fail.
+        """
+        return self._wrapped_loop.__class__
```
clokep, line 18: Is this here to make our own assertion pass or will something deep in asyncio complain without this?

reivilibre: asyncio's `set_event_loop` will complain without this.

tests/asyncio_test_helpers.py
reivilibre: At first I would have sait not; `ContextVar` didn't exist before Py 3.7.
But Sygnal is Py 3.7+ anyway, so lucky me :).

reivilibre:
> it seems like most of the comments in test_httpproxy_asyncio.py weren't handled yet!

Brk, sorry. Didn't notice GitHub had 'minified' them >.<.

tests/test_httpproxy_asyncio.py
reivilibre: Added a comment — we will treat these credentials as invalid for the sake of the test.

tests/test_httpproxy_asyncio.py
reivilibre: Have changed to 
```
        # *now* we should have switched over from the HTTP CONNECT protocol
        # to the user protocol (in our case, a MockProtocol).
```

does this make sense?

tests/test_httpproxy_asyncio.py
```diff
@@ -170,15 +173,16 @@ def test_connect_failure(self):
...
+        # send a sad response with a HTML error page
         fake_proxy.pretend_to_receive(
             b"HTTP/1.0 401 Unauthorised\r\n\r\n<HTML>... some error here ...</HTML>"
         )
 
         # advance event loop because we have to let coroutines be executed
         self.loop.advance(1.0)
 
-        # *now* we should be complete
+        # *now* this future should have completed (
```
clokep, line 185: I think the open paren is a typo?

reivilibre: Yes, sadly I make a few of those random ones somehow. Suspect I had a different window in mind to type to but didn't focus it.

reivilibre:
Thanks for your reviews!

PR #154, opened 2020-10-07
Add platform troubleshooting documentation and sample configuration details

anoadragon453:
A customer had an issue where they were using sandbox push tokens with the production server. I don't blame them, as this doesn't seem to have been documented anywhere.

I've added the `platform` option to the sample config with some explanation, as well as a section in troubleshooting in case others fall down the same path.

While adding to the sample config, I realised that we would eventually end up duplicating nearly every option if we didn't deduplicate early, so that's what I did.

Best reviewed commit-by-commit.

### DIFF
--- a/changelog.d/154.doc
+++ b/changelog.d/154.doc
@@ -1,0 +1,1 @@
+Add troubleshooting documentation and sample configuration about APNs pushkin's `platform` option. 
--- a/changelog.d/154.doc
+++ b/changelog.d/154.doc
@@ -1,1 +1,1 @@
-Add troubleshooting documentation and sample configuration about APNs pushkin's `platform` option. 
+Add troubleshooting documentation and sample configuration about APNs pushkin's `platform` option.
--- a/sygnal.yaml.sample
+++ b/sygnal.yaml.sample
@@ -213,6 +213,16 @@ apps:
   #  # Defaults to 512 if unset.
   #  #
   #  #inflight_request_limit: 512
+  #
+  #  # Specifies whether to use the production or sandbox APNs server. Note that
+  #  # sandbox tokens should only be used with the sandbox server and vice versa. 
+  #  #
+  #  # Valid options are:
+  #  #   * production
+  #  #   * sandbox
+  #  #
+  #  # The default is 'production'.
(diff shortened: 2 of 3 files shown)


sygnal.yaml.sample
```diff
@@ -231,12 +225,23 @@ apps:
...
+  #
+  #  # Specifies whether to use the production or sandbox APNs server. Note that
+  #  # sandbox tokens should only be used with the sandbox server and vice versa. 
+  #  #
+  #  # Valid options are:
+  #  #   * production
+  #  #   * sandbox
+  #  #
+  #  # The default is 'production'.
+  #  #platform: sandbox
```
clokep, line 237: Maybe we want to start making this like the synapse one where we say "Uncomment the following to use sandbox mode" or whatever?

changelog.d/154.doc
```diff
@@ -0,0 +1 @@
+Add troubleshooting documentation and sample configuration about APNs pushkin's `platform` option.
```
clokep, line 1:
```suggestion
Add troubleshooting documentation and sample configuration about APNs pushkin's `platform` option.
```

clokep (approved):
Looks good!

The sample config is confusing with the multiple layers of comments, but I don't want to ask you to fix that here. 😄

sygnal.yaml.sample
```diff
@@ -199,30 +199,24 @@ metrics:
 # For the type, you may specify a fully-qualified Python classname if desired.
 #
 apps:
-  # This is an example APNs push configuration using certificate authentication.
+  # This is an example APNs push configuration
   #
-  #com.example.myapp.ios:
+  #com.example.myapp2.ios:
```
clokep, line 204: Any particular reason for the `myapp2` here? Seems confusing if there's no `myapp`?

anoadragon453: Oops, simply because there used to be a `myapp` and `myapp2` before the deduplication.

sygnal.yaml.sample
anoadragon453: I am definitely not opposed to that!

PR #155, opened 2020-10-07
Add note about warning the world before making a release

anoadragon453:
Otherwise we end up making many releases without anybody knowing.


clokep (approved):

clokep:
Although looks like some CI magic is needed? Might just be the newsfile though.

PR #158, opened 2020-11-02
Update troubleshooting document with 400 BadDeviceToken

reivilibre:
Someone came in the room with this issue a few days ago, thought I'd document it.

But then found that Andrew already did, but not in the right section.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -126,7 +126,20 @@ Instead of configuring Sygnal with your APNs secrets, you need to configure
 Firebase with your APNs secrets, and Sygnal with your Firebase secrets.
 
 
-#### Common cause 3: Using sandbox tokens with the production APNs server
+### App doesn't receive notifications when inactive
+
+If you want your application to be woken up to be able to process APNs messages
+received when your application is in the background, you need to set the
+`content-available` flag in your pusher's default payload — see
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -126,7 +126,20 @@ Instead of configuring Sygnal with your APNs secrets, you need to configure
...
+If you want your application to be woken up to be able to process APNs messages
+received when your application is in the background, you need to set the
+`content-available` flag in your pusher's default payload — see
+[the notes for iOS applications](applications.md#ios-applications-beware).
+
+
+### '400 BadDeviceToken' error
+
+If you have a bad device token and your token was not a fictitious invention,
+the most likely cause is that you have used a token from the wrong 'environment',
```
erikjohnston, line 140:
```suggestion
If you get a bad device token error the an possible cause is that
you have used a token from the wrong 'environment',
```

anoadragon453: > error the an possible

extra "an"

anoadragon453 (approved):

docs/troubleshooting.md
anoadragon453, line 140:
```suggestion
If you get a bad device token error and you have doubled-checked the
token is correct, it is possible that you have used a token from the wrong 'environment',
```

Issue #144, opened 2020-08-05
Sygnal does not recover cleanly if it builds a queue of messages


michaelkaye:
By inspection, sygnal runs into problems when it builds a queue of messages (for whatever reason).

We should more aggressively shed the load queued messages to stop this needing manual intervention.

The flow as I see it from an incident today:

t=0 - high response times from remote server causing a queue to build up
t=10 - response times drop to normal levels, queue has built up to > 2k items
t=15 - queue continues to build but response times are back to that earlier
t=20 - restart sygnal - backlog of requests is dropped, retries cause another spike in load and increase in response time, but no queue builds up. Once messages are cleared all metrics return to normal.

It seems to me that the queue itself is causing the performance issues. A request that is more than 60s old is likely to have been thrown away by whatever http request servlets are in the flow, or rejected by a loadbalancer or proxy, so we should discard them and not attempt to process them.

erikjohnston:
This is for APNs. Have some graphs:

![Selection_026](https://user-images.githubusercontent.com/8428120/89510525-94e9c200-d7c8-11ea-8f27-c783bb8ac70a.png)

(This is https://grafana.matrix.org/d/rg3tCbWWk/sygnal?orgId=1&from=1596660536006&to=1596661385527&var-bucket_size=$__auto_interval_bucket_size&var-instance=corus.priv.matrix.org:7564 for those with access)

erikjohnston:
We are running v0.7.1, which is out of date. Let's update it.

michaelkaye:
Looking better now; we're cleanly recovering for this failure.

It still causes up to 50% of pushes to fail back to synapse during this period:

![only_one_pusher](https://user-images.githubusercontent.com/1917473/89640099-c5079280-d8a6-11ea-9169-5f54dd9fb1ae.png)
![snappy](https://user-images.githubusercontent.com/1917473/89640100-c5a02900-d8a6-11ea-95ff-4175209b0730.png)

Interestingly, this is only happening on one of the two synapse pushers - perhaps something to investigate there.

michaelkaye:
(times are off because one is UTC one is BST)

erikjohnston:
One potential cause to look into is if we're correctly timing out connections/requests. If we have a fixed connection pool then its possible all connections get stuck for an age.

michaelkaye:
This failed again today, and seems to have failed earlier in the week as well.

[...]

PR #163, opened 2021-01-17
Example - Optionally exit python if limit hit.

michaelkaye:
If the system becomes unstable, then exit rather than continuing - this will allow something like systemd to restart quickly and let us carry on from a known-good situation.


michaelkaye:
hides impact of #144

PR #172, opened 2021-03-02
Initialise sygnal_inflight_request_limit_drop before it's incremented

anoadragon453:
The `sygnal_inflight_request_limit_drop` is a Counter that counts how many requests were dropped due to a pushkin hitting its concurrent request limit, and was added in https://github.com/matrix-org/sygnal/pull/146.

A quirk of counter metrics with a label is that they will not show up in prometheus until they are initially referenced. This had the side-effect of not being able to easily monitor/alert on the metric until something actually went wrong

This PR references and sets up a instance variable with the pushkin's name as a label immediately, ensuring it appears as 0 in prometheus upon startup.

### DIFF
--- a/changelog.d/172.bugfix
+++ b/changelog.d/172.bugfix
@@ -1,0 +1,1 @@
+Fixes a bug where the `sygnal_inflight_request_limit_drop` metric would not appear in prometheus until requests were actually dropped.
--- a/sygnal/helper/proxy/proxy_asyncio.py
+++ b/sygnal/helper/proxy/proxy_asyncio.py
@@ -260,9 +260,7 @@ class ProxyingEventLoopWrapper:
     """
 
     def __init__(
-        self,
-        wrapped_loop: asyncio.AbstractEventLoop,
-        proxy_url_str: str,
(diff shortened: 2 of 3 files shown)


clokep:
I think this just needs to run `black .`?

clokep:
I think black and flake8 are fighting? :) Might want to check our synapse config for flake8 and copy it over.

anoadragon453:
> I think black and flake8 are fighting? :) Might want to check our synapse config for flake8 and copy it over.

I'll do that in a separate PR. This one should be good to review now.

Interestingly in Synapse we [disable](https://github.com/matrix-org/synapse/blob/7b06f85c0e18b62775f12789fdf4adb6a0a47a4b/setup.cfg#L20) long line checks in flake8, though black seemed perfectly happy to include a 100+ char line. So there's probably some `black` config somewhere (other than [here](https://github.com/matrix-org/synapse/blob/7b06f85c0e18b62775f12789fdf4adb6a0a47a4b/pyproject.toml#L37-L56)?) that sets a line length limit.

clokep (approved):

PR #174, opened 2021-03-05
Remove useless log line

anoadragon453:
I initially tried adapting this log line to include some information relevant to the request, but all that's available here for the device was its app_id (too generic) and pushkey (probably not a good idea to log).

So instead I've opted to remove it as I feel it's not too useful on its own (it's also quite spammy). Plus you probably only care when a push notification fails to send, which is logged below.

### DIFF
--- a/changelog.d/174.misc
+++ b/changelog.d/174.misc
@@ -1,0 +1,1 @@
+Remove extraneous log line.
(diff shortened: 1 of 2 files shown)


clokep:
@anoadragon453 I think this just needs a merge in so CI passes? This looks fine to me.

clokep (approved):

Issue #176, opened 2021-03-08
Web Push support


bwindels:
To support push notification on hydrogen, it would be valuable to have web push support in sygnal

bwindels:
Potential library to use: https://github.com/web-push-libs/pywebpush

richvdh:
I think this is the same as https://github.com/matrix-org/synapse/issues/1666.

PR #177, opened 2021-03-11
Add experimental support for WebPush

bwindels:
#176 

~~Needs a change in a dependency so had to fork it for now, have a PR to merge it upstream at https://github.com/web-push-libs/pywebpush/pull/133~~ This has been merged and released upstream.


sygnal/webpushpushkin.py
```diff
@@ -0,0 +1,167 @@
...
+        "max_connections",
+        "vapid_private_key",
+        "vapid_contact_email",
+    } | ConcurrencyLimitedPushkin.UNDERSTOOD_CONFIG_FIELDS
+
+    def __init__(self, name, sygnal, config):
+        super(WebpushPushkin, self).__init__(name, sygnal, config)
+
+        nonunderstood = set(self.cfg.keys()).difference(self.UNDERSTOOD_CONFIG_FIELDS)
+        if len(nonunderstood) > 0:
```
callahad, line 66: Might be more clearly expressed as:

```python
nonunderstood = self.cfg.keys() - self.UNDERSTOOD_CONFIG_FIELDS
if nonunderstood:
    ...
```

But these follow patterns elsewhere in Sygnal so 🤷

sygnal/webpushpushkin.py
```diff
@@ -0,0 +1,167 @@
...
+            pool=self.http_pool,
+            contextFactory=tls_client_options_factory,
+            proxy_url_str=proxy_url,
+        )
+        self.http_agent_wrapper = HttpAgentWrapper(self.http_agent)
+
+        privkey_filename = self.get_config("vapid_private_key")
+        self.vapid_private_key = Vapid.from_file(private_key_file=self.get_config("vapid_private_key"))
+        vapid_contact_email = self.get_config("vapid_contact_email")
+        self.vapid_claims = {"sub": "mailto:{}".format(vapid_contact_email)}
```
callahad, line 95: Will this blow up if the VAPID configuration isn't present? Should we handle that better?

sygnal/webpushpushkin.py
```diff
@@ -0,0 +1,167 @@
...
+            'prio': n.prio,
+            'membership': n.membership,
+            'sender_display_name': n.sender_display_name,
+            'event_id': n.event_id,
+            'room_id': n.room_id,
+            'user_is_target': n.user_is_target,
+            'type': n.type,
+            'sender': n.sender,
+            'session_id': session_id,
+        }
```
callahad, line 121: What happens if some of these values are missing from `n`?

callahad (dismissed):

sygnal/webpushpushkin.py
bwindels: Yeah, still need to work on the format of the actual payload :+1:

sygnal/webpushpushkin.py
bwindels: indeed, I assume by throwing a `PushkinSetupException` if unhappy

setup.py
```diff
@@ -49,6 +49,8 @@ def read(fname):
         "idna>=2.8",
         "psycopg2>=2.8.4",
         "importlib_metadata",
+        "pywebpush@git+https://github.com/matrix-org/pywebpush"
```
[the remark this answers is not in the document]
callahad: Using:

`pywebpush @ https://github.com/matrix-org/pywebpush/archive/HASH_HERE.tar.gz`

would probably be a little quicker than a full clone of the repo

bwindels:
Tried to address as many outstanding issues as I could. A few things to note:
 - I've flat out copied the histogram/gauge code for connection handling from the GCM pushkin, and don't really know what I'm doing there, so perhaps a detailed look would be good
 - Do I need to add any specific opentracing logging?
 - I could use some help solving the typing errors. It seems to give an error on the libraries I added: `pywebpush` and `py_vapid`. Not sure how to solve that.

bwindels:
Should I ideally write any tests for the new pushkin?

sygnal/webpushpushkin.py
```diff
@@ -0,0 +1,215 @@
...
+            device (Device): Device information to which the constructed payload
+            will be sent.
+
+        Returns:
+            JSON-compatible dict
+        """
+        payload = {}
+
+        if device.data:
+            payload.update(device.data.get("default_payload", {}))
```
bwindels, line 163: Copied this over from GCM, but wondering what it would do if default_payload is set to something that is not a dictionary, like `null`?

clokep: In both cases we should probably either assert the type or do `data.get("default_payload") or {}` (which means any false-y value will be replaced with `{}`). This would still break if `default_payload` was `"foo"` or whatever...

bwindels: Added better validate for the push data. Shall I change GCM to do the same in this PR as well or a different one?

docs/applications.md
```diff
@@ -203,3 +203,42 @@ within FCM's limit.
 Please also note that some fields will be unavailable if you registered a pusher
 with `event_id_only` format.
 
+### Web Push
+
+#### Setup & configuration
+
+In the sygnal virtualenv, generate the server key pair by running `vapid --gen --applicationServerKey`. This will generate a `private_key.pem` (which you'll refer to in the config file with `vapid_private_key`) and `public_key.pem` file, and also string labeled `Application Server Key`. You'll copy the Application Server Key to your web application to subscribe to the push manager:
```
clokep, line 210: Can you wrap each of these paragraphs? Thanks!

docs/applications.md
```diff
@@ -203,3 +203,42 @@ within FCM's limit.
...
+
+In your web application, [the push manager subscribe method](https://developer.mozilla.org/en-US/docs/Web/API/PushManager/subscribe) will return [a subscription](https://developer.mozilla.org/en-US/docs/Web/API/PushSubscription) with an `endpoint` and `keys` property, the latter containing a `p256dh` and `auth` property. The `p256dh` key is used as the push key, and the push data is expected `endpoint` and `auth`. You can also set `default_payload` in the push data; any properties set in it will be present in the push messages you receive, so it can be used to pass identifiers specific to your client (like which account the notification is for).
+
+Also note that because you can only have one push subscription per service worker, and hence per origin, you might create pushers for different accounts with the same p256dh push key. To prevent the server from removing other pushers with the same push key for your other users, you should set `append` to `true` when uploading your pusher.
+
+#### Notification format
+
+The notification as received by your web application will contain these keys if they were set by the homeserver, and otherwise omit them.
```
clokep, line 229: This feels like it should link to the spec somewhere instead of listing them out again?

setup.py
```diff
@@ -49,6 +49,8 @@ def read(fname):
         "idna>=2.8",
         "psycopg2>=2.8.4",
         "importlib_metadata",
+        "pywebpush>=1.13.0",
+        "py-vapid>=1.7.0",
```
clokep, line 53: As someone who has no added libraries before, do we care about the licenses of these or anything of that nature? (Probably a question for @matrix-org/synapse-core)

sygnal/webpushpushkin.py
```diff
@@ -0,0 +1,225 @@
...
+            proxy_url_str=proxy_url,
+        )
+        self.http_agent_wrapper = HttpAgentWrapper(self.http_agent)
+
+        privkey_filename = self.get_config("vapid_private_key")
+        if not privkey_filename:
+            raise PushkinSetupException("'vapid_private_key' not set in config")
+        if not os.path.exists(privkey_filename):
+            raise PushkinSetupException("path in 'vapid_private_key' does not exist")
+        self.vapid_private_key = Vapid.from_file(private_key_file=privkey_filename)
```
clokep, line 103: Can this raise any exceptions? If so we'll want to catch them and `raise PushkinSetupException(...) from e`

sygnal/webpushpushkin.py
```diff
@@ -0,0 +1,225 @@
...
+        vapid_contact_email = self.get_config("vapid_contact_email")
+        if not vapid_contact_email:
+            raise PushkinSetupException("'vapid_contact_email' not set in config")
+        self.vapid_claims = {"sub": "mailto:{}".format(vapid_contact_email)}
+
+    async def _dispatch_notification_unlimited(self, n, device, context):
+        p256dh = device.pushkey
+        if not isinstance(device.data, dict):
+            logger.info("device.data is not a dict, reject pushkey")
```
clokep, line 112: This should probably be a warning (or maybe an error) log?

Probably the same for the other logging below.

Also should these log lines include some info about the device or something so it can be traced?

sygnal/webpushpushkin.py
```diff
@@ -0,0 +1,225 @@
...
+            "membership",
+            "event_id",
+            "sender",
+            "sender_display_name",
+            "user_is_target",
+            "type",
+            "content",
+        ]:
+            if getattr(n, attr, None):
+                payload[attr] = getattr(n, attr)
```
clokep, line 188: Instead of calling getattr a few times, this is a bit simpler.

```suggestion
            value = getattr(n, attr, None)
            if value:
                payload[attr] = value
```

sygnal/webpushpushkin.py
```diff
@@ -0,0 +1,225 @@
...
+
+        if getattr(n, "counts", None):
+            counts = n.counts
```
clokep, line 191:
```suggestion
        counts = getattr(n, "count", None)
        if count is not None:
```

sygnal/webpushpushkin.py
```diff
@@ -0,0 +1,225 @@
...
+            for attr in ["unread", "missed_calls"]:
+                if getattr(counts, attr, None) is not None:
+                    payload[attr] = getattr(counts, attr)
+
+        return payload
+
+
+class HttpAgentWrapper:
```
clokep, line 199: Can we add a docstring here saying what this is for? It seems to:

* Provide a `post` method that matches the API expected from pywebpush.

Maybe it does more though?

sygnal/webpushpushkin.py
```diff
@@ -0,0 +1,225 @@
...
+        deferred = self.http_agent.request(
+            b"POST",
+            endpoint.encode(),
+            headers=Headers(headers),
+            bodyProducer=body_producer,
+        )
+        return HttpResponseWrapper(deferred)
+
+
+class HttpResponseWrapper:
```
clokep, line 221: This could also use a docstring, it seems to provide a status code < 202 to avoid error checking in in pywebpush.

I'm not sure why text is provided?

sygnal/webpushpushkin.py
```diff
@@ -0,0 +1,225 @@
...
+                    payload[attr] = getattr(counts, attr)
+
+        return payload
+
+
+class HttpAgentWrapper:
+    def __init__(self, http_agent):
+        self.http_agent = http_agent
+
+    def post(self, endpoint, data, headers, timeout):
```
clokep, line 203: Can you add a docstring about what the inputs are here?

clokep (changes requested):
Looks pretty reasonable all-in-all! I left a few comments, but they're mostly "please add comments". 😄

setup.py
bwindels: [py-vapid](https://pypi.org/project/py-vapid/) and [pywebpush](https://pypi.org/project/pywebpush/) are both MPL2. Should transitive dependencies be checked?

sygnal/webpushpushkin.py
bwindels: text is references by [pywebpush when it logs the response](https://github.com/web-push-libs/pywebpush/blob/main/pywebpush/__init__.py#L373).

sygnal/webpushpushkin.py
bwindels: No, that's it. Added docstrings :+1:

docs/applications.md
bwindels: I've added them here again since every pushkin seems to have their own format they push to their clients, and the format is slightly different than the spec (counts is flattened). I've mentioned this in the docs now and linked to the spec.

sygnal/webpushpushkin.py
clokep: Oh I missed this. I think while we're here we should do it. 👍

sygnal/webpushpushkin.py
clokep: Thanks! I missed that. 😄

clokep:
@bwindels Can you double check over the changes I pushed and make sure they seem reasonable?

If so I think this is ✅ !

docs/applications.md
```diff
@@ -246,11 +246,11 @@ your pusher.
 
 #### Notification format
 
-The notification as received by your web application will contain these keys 
-if they were set by the homeserver, and otherwise omit them. These are the
-same as specified in [the push gateway spec]
-(https://matrix.org/docs/spec/push_gateway/r0.1.0#post-matrix-push-v1-notify),
-but `counts` with `unread` and `missed_calls` is flattened into the notification object.
+The notification as received by your web application will contain the following keys 
+(assuming they were sent by the homeserver). These are the
```
bwindels, line 250: I meant more that if they are null in the notification from the homeserver, they will be omitted in the payload to the push gateway. Not sure the distinction is important though.

bwindels:
Thank you for your changes @clokep! Lgtm (one minor comment).

docs/applications.md
clokep: Ah, I misunderstood this then. I don't know if that distinction is important, yeah. I think the important bit is that you should handle them not being there?

docs/applications.md
```diff
@@ -203,3 +203,66 @@ within FCM's limit.
...
+Also note that because you can only have one push subscription per service worker,
+and hence per origin, you might create pushers for different accounts with the same 
+p256dh push key. To prevent the server from removing other pushers with the same 
+push key for your other users, you should set `append` to `true` when uploading 
+your pusher.
+
+#### Notification format
+
+The notification as received by your web application will contain the following keys 
+(assuming they were sent by the homeserver). These are the
```
clokep, line 250:
```suggestion
The notification as received by your web application will contain the following keys 
(assuming non-null values were sent by the homeserver). These are the
```

clokep (approved):

sygnal/webpushpushkin.py
```diff
@@ -102,7 +102,7 @@ def __init__(self, name, sygnal, config):
             raise PushkinSetupException("path in 'vapid_private_key' does not exist")
         try:
             self.vapid_private_key = Vapid.from_file(private_key_file=privkey_filename)
-        except BaseException as e:
```
bwindels: I had this first fwiw, but switched to the more general one because I assumed [open and read here](https://github.com/web-push-libs/vapid/blob/main/python/py_vapid/__init__.py#L118) could throw some kind of I/O exception?

bwindels: Ah, it's merged already :tada:  probably not too important in any case. Thanks for seeing this through!

clokep: > I had this first fwiw, but switched to the more general one because I assumed [open and read here](https://github.com/web-push-libs/vapid/blob/main/python/py_vapid/__init__.py#L118-L119) could throw some kind of I/O exception?

They could, but catching `BaseExeption` is usually not great. We check that the file exists above so I suspect that wouldn't happen...although I guess there could be permission issues.

PR #180, opened 2021-03-19
Fix webpush vapid_claims from one request bleeding into others

bwindels:


bwindels:
Having a look if I can add tests ...

bwindels:
We need more infrastructure to be able to write unit tests somewhat easily, so parked that work in the `bwindels/webpush-unit-tests` branch and leaving the tests out for now.

sygnal/webpushpushkin.py
```diff
@@ -272,3 +296,6 @@ class HttpResponseWrapper:
 
     def __init__(self, deferred):
         self.deferred = deferred
+
+    async def read_body(self, response):
+        return (await readBody(response)).decode()
```
clokep, line 301: I'm not sure this is really useful to have in the wrapper since `pywebpush` doesn't expect this? It seems to be conflating concerns. Can we inline this code above?

sygnal/webpushpushkin.py
```diff
@@ -137,6 +138,11 @@ async def _dispatch_notification_unlimited(self, n, device, context):
         payload = WebpushPushkin._build_payload(n, device)
         data = json.dumps(payload)
 
+        # note that webpush modifies vapid_claims, so make sure it's only used once
+        vapid_claims = {
+            "sub": "mailto:{}".format(self.vapid_contact_email),
+            "exp": self._get_vapid_exp(),
```
clokep, line 144: This seems fine, but I do wonder if we should leave manually specifying `exp` out until we get the tests going as it isn't necessary to copy this out of pywebpush yet.

changelog.d/180.bugfix
```diff
@@ -0,0 +1,8 @@
+Fix vapid_claims from one webpush request bleeding into others
+
+The vapid_claims dict was reused across requests, but pywebpush actually modifies this dictionary.
+This made the pushkin usable with only a single pusher for a single user, any other pushers
+would be rejected by their gateway, at least if from a different provider, as the aud claim
+(the origin of the endpoint url) didn't match the server name, but would likely even fail for 
+a single provider as the expiry time wasn't refreshed.
+The resut was the pushkey would be rejected, removing the pusher.
```
clokep, line 8: And this should only be a single line. I think the top line is OK?

clokep (approved):
LGTM! Thanks!

Issue #167, opened 2021-02-19
Using a HTTP proxy sometimes breaks TLS connections


erikjohnston:
Stack trace, with (I think) related log lines, though there are multiple HTTP connections happening at once:

```
2021-02-16 08:51:51,072 [14936] DEBUG sygnal.helper.proxy.proxy_asyncio Initiating proxy CONNECT
2021-02-16 08:51:51,092 [14936] DEBUG sygnal.helper.proxy.proxy_asyncio CONNECT response from proxy: hv=b'HTTP/1.0', r=b'200', rp=b'Connection established'
2021-02-16 08:51:51,092 [14936] DEBUG sygnal.helper.proxy.proxy_asyncio Ready to switch over protocol
2021-02-16 08:51:51,679 [14936] DEBUG aioapns Remote setting changed: ChangedSetting(setting=SettingCodes.HEADER_TABLE_SIZE, original_value=4096, new_value=4096)
2021-02-16 08:51:51,680 [14936] DEBUG aioapns Remote setting changed: ChangedSetting(setting=SettingCodes.MAX_CONCURRENT_STREAMS, original_value=None, new_value=1)
2021-02-16 08:51:51,680 [14936] DEBUG aioapns Remote setting changed: ChangedSetting(setting=SettingCodes.MAX_FRAME_SIZE, original_value=16384, new_value=16384)
2021-02-16 08:51:51,680 [14936] DEBUG aioapns Remote setting changed: ChangedSetting(setting=SettingCodes.MAX_HEADER_LIST_SIZE, original_value=None, new_value=8000)
2021-02-16 08:51:51,680 [14936] ERROR asyncio application protocol failed to receive SSL data
protocol: <asyncio.sslproto.SSLProtocol object at 0x7f54156d1d50>
transport: <_SelectorSocketTransport fd=23 read=polling write=<idle, bufsize=0>>
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/asyncio/sslproto.py", line 545, in data_received
    self._app_protocol.data_received(chunk)
  File "/opt/matrix/sygnal/lib/python3.7/site-packages/aioapns/connection.py", line 100, in data_received
    self.flush()
  File "/opt/matrix/sygnal/lib/python3.7/site-packages/aioapns/connection.py", line 199, in flush
    self.transport.write(self.conn.data_to_send())
AttributeError: 'NoneType' object has no attribute 'write'
```

[...]

PR #181, opened 2021-03-19
Fix race when using HTTP proxy.

erikjohnston:
After `start_tls` we need to manually call `connection_made` on the protocol to tell it about the new underlying transport. This gives a brief window where the protocol can receive data without having a transport to write to, causing issues with the APNS connections where it assumes it can write once it starts reading data.

We fix this by wrapping the protocol in a buffer that simply buffers incoming data until `connection_made` is called.

Fixes https://github.com/matrix-org/sygnal/issues/167, hopefully

### DIFF
--- a/tests/asyncio_test_helpers.py
+++ b/tests/asyncio_test_helpers.py
@@ -70,7 +70,7 @@ class TimelessEventLoopWrapper:
     ):
         self.call_at(self._time + delay, callback, *args, context=context)
 
-        # We're mean to return a canceller, but can cheat and return a no-op one
+        # We're meant to return a canceller, but can cheat and return a no-op one
         # instead.
         class _Canceller:
             def cancel(self):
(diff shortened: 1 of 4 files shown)


tests/asyncio_test_helpers.py
```diff
@@ -70,6 +70,14 @@ def call_later(
     ):
         self.call_at(self._time + delay, callback, *args, context=context)
 
+        # We're mean to return a canceller, but can cheat and return a no-op one
+        # instead.
```
clokep, line 74:
```suggestion
        # We're meant to return a canceller, but can cheat and return a no-op one
        # instead.
```

clokep (approved):
I don't really see how this is happening (maybe the Twisted reactor is stealing control during the await?) Seems like this should fix the issue though.

erikjohnston:
> I don't really see how this is happening (maybe the Twisted reactor is stealing control during the await?) Seems like this should fix the issue though.

It's possible there's just a lot of work queued up, or a GC happens, or some TLS session resumption that allows the server to send data immediately, or... something. I don't really know :/

PR #182, opened 2021-03-22
Whitelist webpush endpoints

bwindels:
through an optional `allowed_endpoints` configuration key

I added support for globs because some browser vendor endpoints seem to contain sharding identifiers in the subdomain.

### DIFF
--- a/changelog.d/182.feature
+++ b/changelog.d/182.feature
@@ -1,1 +1,1 @@
-Add 'allowed_endpoints' configuration option for webpush pushkins to whitelist the allowed endpoints.
+Add 'allowed_endpoints' configuration option for limiting the endpoints that WebPush pushkins will contact.
--- a/changelog.d/182.feature
+++ b/changelog.d/182.feature
@@ -1,0 +1,1 @@
+Add 'allowed_endpoints' configuration option for webpush pushkins to whitelist the allowed endpoints.
--- a/sygnal/webpushpushkin.py
+++ b/sygnal/webpushpushkin.py
@@ -137,8 +137,8 @@ class WebpushPushkin(ConcurrencyLimitedPushkin):
                     "push gateway %s is not in allowed_endpoints, blocking request",
                     endpoint_domain,
                 )
+                # abort, but don't reject push key
                 return []
-                # don't reject push key though
 
         if not p256dh or not endpoint or not auth:
             logger.warn(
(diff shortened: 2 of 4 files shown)


sygnal/webpushpushkin.py
```diff
@@ -104,26 +113,42 @@ def __init__(self, name, sygnal, config):
...
+        endpoint_domain = urlparse(endpoint).netloc
+        if self.allowed_endpoints:
+            match = next(
+                (
+                    regex
+                    for regex in self.allowed_endpoints
+                    if regex.fullmatch(endpoint_domain)
+                ),
+                None,
+            )
```
clokep, line 139: This is pretty difficult to read, I think you want something more like:

```python
allowed = any(regex.fullmatch(endpoint_domain) for regex in self.allowed_endpoints)
```

clokep (changes requested):
Looks good overall.

clokep:
Does it make sense to include some info on Google/Microsoft/Mozilla here so people know what a reasonable value is?

sygnal/webpushpushkin.py
bwindels: that is way better indeed, only came up with that monstrosity after looking online how to to do a `find`

bwindels:
> Does it make sense to include some info on Google/Microsoft/Mozilla here so people know what a reasonable value is?

Added this to the docs :+1:

bwindels:
How would I fix this error?:
```
sygnal/webpushpushkin.py:108: error: Incompatible types in assignment (expression has type "None", variable has type "List[Any]")  [assignment]
```

clokep:
> How would I fix this error?:
> 
> ```
> sygnal/webpushpushkin.py:108: error: Incompatible types in assignment (expression has type "None", variable has type "List[Any]")  [assignment]
> ```

When you define `self.allowed_endpoints` you need to add a type to it. I usually do this as:

```python
self.foo = None  # type: Optional[List[str]]
if blah:
    self.foo = list(...)
```

You'll also need a `from typing import List, Optional` at the top.

changelog.d/182.feature
```diff
@@ -0,0 +1 @@
+Add 'allowed_endpoints' configuration option for webpush pushkins to whitelist the allowed endpoints.
```
clokep, line 1:
```suggestion
Add 'allowed_endpoints' configuration option for limiting the endpoints that WebPush pushkins will contact.
```

clokep (commented):
Looks good! I think my other comment will fix the mypy issue.

clokep (approved):
LGTM!

PR #184, opened 2021-03-23
add allowed_endpoints option as understood config option

bwindels:


changelog.d/184.bugfix
```diff
@@ -0,0 +1 @@
+add allowed_endpoints option as understood config option
```
clokep, line 1:
```suggestion
Add `allowed_endpoints` option as understood config option for WebPush pushkins.
```

PR #185, opened 2021-03-26
set time-to-live for webpush messages

bwindels:
Configurable with a default of 15 minutes.

### DIFF
--- a/sygnal/webpushpushkin.py
+++ b/sygnal/webpushpushkin.py
@@ -121,12 +121,9 @@ class WebpushPushkin(ConcurrencyLimitedPushkin):
         self.vapid_contact_email = self.get_config("vapid_contact_email")
         if not self.vapid_contact_email:
             raise PushkinSetupException("'vapid_contact_email' not set in config")
-        self.ttl = self.get_config("ttl")
-        if self.ttl:
-            if not isinstance(self.ttl, int):
-                raise PushkinSetupException("'ttl' must be an int if set")
-        else:
-            self.ttl = DEFAULT_TTL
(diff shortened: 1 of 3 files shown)


sygnal/webpushpushkin.py
```diff
@@ -119,6 +120,12 @@ def __init__(self, name, sygnal, config):
         self.vapid_contact_email = self.get_config("vapid_contact_email")
         if not self.vapid_contact_email:
             raise PushkinSetupException("'vapid_contact_email' not set in config")
+        self.ttl = self.get_config("ttl")
+        if self.ttl:
+            if not isinstance(self.ttl, int):
+                raise PushkinSetupException("'ttl' must be an int if set")
+        else:
+            self.ttl = DEFAULT_TTL
```
callahad, line 128: Would it be cleaner to do this?

```python
self.ttl = self.get_config("ttl", DEFAULT_TTL)
if not isinstance(self.ttl, int):
    raise PushkinSetupException("'ttl' must be an int if set")
```

callahad:
Do we need to document the `ttl` configuration anywhere? (The VAPID config isn't documented either, so I guess we can let this slip, too, while WebPush is experimental)

Does `ttl` need to be added to the `UNDERSTOOD_CONFIG_FIELDS` list?

bwindels:
Thanks for the review @callahad, ready for another look.

callahad:
r+ once CI passes and if you promise that you've tested this locally and it at least appears to work.

callahad (approved):

bwindels:
@callahad Have tested again, all seems to work :+1:

PR #193, opened 2021-03-29
Update Tox settings for easier use & better performance

### LINKED ISSUE #189: Tox should default to current python, not py37
Right now our tox.ini (and docs) default to `py37`... we should replace this with just `py` which means to use the same version of Python which invoked tox.

This would make testing more convenient for folks on py38, py39, etc.

### LINKED ISSUE #191: Ignore virtualenvs when running lints (flake8, black, etc) with tox
Currently our CONTRIBUTING docs recommend creating a virtualenv in the same folder as Sygnal, but our tox.ini defaults to running linters over the current directory (e.g., `flake8 .`)... which then tries to apply the linter to _everything_ in the current directory, including your virtualenv.

We should probably edit tox.ini to only lint `sygnal/` `tests/` and `setup.py`

While we're at it, we should look into our `scripts-dev/lint.sh` and determine if that's used anywhere or if it's redundant with `tox`

aitorres:
## What's changed

This PR updates Tox's settings in order to:

- Consider only paths and files related to the project while running code style / lint / type checks, instead of the whole project (which could potentially include a virtual environment) (as reported in #191).
- Remove the `py37` constraint while setting up the environment, in order to let users run Tox with other versions (specifically newer versions of Python) (as reported in #189)

CONTRIBUTING.md was also updated to remove references to the `lint.sh` script, which was removed, and replace its usage with `tox` in order to achieve the same effects.

Two entries were added to the changelog, for `doc` and `misc` changes.

## Solved issues

Closes #189 if merged.
Closes #191 if merged.


CONTRIBUTING.md
```diff
@@ -88,27 +88,25 @@ Sygnal follows the [Synapse code style].
...
 may wish to ensure any new code is committed first**.
 
-By default, this script checks all files and can take some time; if you alter
-only certain files, you might wish to specify paths as arguments to reduce the
-run-time:
-
-```
-./scripts-dev/lint.sh path/to/file1.py path/to/file2.py path/to/folder
-```
+These commands will consider the paths and files related to the project (i.e. everything in `sygnal/` and in `tests/` as well as the `setup.py` file).
```
callahad, line 109: I'm not sure _why_ we do this, but it seems like the rest of the file has manual line-wrapping at ~78 characters... Could you please break this line to match the rest?

CONTRIBUTING.md
```diff
@@ -88,27 +88,25 @@ Sygnal follows the [Synapse code style].
...
+# Activate the virtual environment
+source ./venv/bin/activate
 
-# Run the linter script
-./scripts-dev/lint.sh
+# Run the code style check
+tox -e check_codestyle
+
+# Run the types check
+tox -e check_codestyle
```
callahad, line 103: I think this one should be `check_types`

callahad (changes requested):
Looks good, thank you for the contribution! Two very small nits and then it should be good to merge.

aitorres:
All done! Those two things flew past my radar 😅 fixed now.

By the way, not sure if this is the most appropriate place for this: I would like to start contributing to Matrix in general with some frequency, but don't know what's the best way to start. Is there any public, central hub / channel for contributors to communicate?

CONTRIBUTING.md
```diff
@@ -88,27 +88,26 @@ Sygnal follows the [Synapse code style].
...
-./scripts-dev/lint.sh
+# Run the code style check
+tox -e check_codestyle
+
+# Run the types check
+tox -e check_types
 ```
 
 **Note that the script does not just test/check, but also reformats code, so you
 may wish to ensure any new code is committed first**.
```
callahad, line 107: We should probably delete this too, since the `tox` commands all have nice `--check-only` flags on them, in contrast to the old `lint.sh`.

Otherwise, this looks good to go from me!

callahad (changes requested):

callahad:
> By the way, not sure if this is the most appropriate place for this: I would like to start contributing to Matrix in general with some frequency, but don't know what's the best way to start. Is there any public, central hub / channel for contributors to communicate?
> 
> Edit: Found the rooms on Matrix :-)

Oh yay! Please do poke around, get a feel for how Matrix works as a platform, etc. and feel free to ask us any questions you might have. I'm usually in the [Synapse Dev](https://matrix.to/#/#synapse-dev:matrix.org) and [Sygnal](https://matrix.to/#/#sygnal:matrix.org) rooms.

callahad (approved):
Hooray! Thank you so much.

PR #194, opened 2021-03-30
fix to "development dependencies should be available as extras"

H-Shay:
Hello-here is my first pull request. Hopefully i hit all the requirements from the 'contributing' doc. 
Signed-off-by: Hillery Shay <shaysquared@gmail.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
         "dev": [
         "black==19.10b0", 
         "flake8==3.8.3", 
-        "isortisort~=5.0", 
+        "isort~=5.0", 
         "mypy==0.780", 
         "mypy-zopemypy-zope==0.2.7", 
         "tox",
--- a/setup.py
+++ b/setup.py
@@ -54,12 +54,12 @@ setup(
     ],
     extras_require={
         "dev": [
-        "black==19.10b0",
-        "flake8==3.8.3",
-        "isort~=5.0",
-        "mypy==0.780",
-        "mypy-zope==0.2.7",
-        "tox",
+            "black==19.10b0",
+            "flake8==3.8.3",
+            "isort~=5.0",
+            "mypy==0.780",
+            "mypy-zope==0.2.7",
+            "tox",
         ]
     },
     long_description=read("README.rst"),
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -52,5 +52,6 @@ def read(fname):
         "pywebpush>=1.13.0",
         "py-vapid>=1.7.0",
     ],
+    extras_require={"dev": ["black", "flake8", "isort", "mypy", "mypy-zope", "tox"]},
```
callahad, line 55: It looks like our [https://github.com/matrix-org/sygnal/blob/master/tox.ini](`tox.ini`) file specifies some specific versions that it expects for these tools; could you please add those in here? (E.g., `"black"` → `"black==19.10b0"`)

callahad (changes requested):
> Hello-here is my first pull request.

Yay! 🎉 Thank you so much! Everything looks in order, but I had a few small changes to request

In addition to the comment above about setting versions in setup.py, could you also please remove any mentions of manually `pip install`ing those tools from CONTRIBUTING.md and change the documented install command to `pip install -e '.[dev]'`, which should pull in Sygnal _and_ the dev dependencies?

clokep:
We should also update `tox.ini` to install this extra instead of separately installing the requirements!

setup.py
H-Shay: Great, I can make these changes sometime tonight. Thanks for the feedback! Will also incorporate the changes from the comment below.

callahad:
> We should also update `tox.ini` to install this extra instead of separately installing the requirements!

Ah, yep! From [the Tox docs](https://tox.readthedocs.io/en/latest/config.html#conf-extras) it looks like we should be able to specify:

```
extras =
  dev
```

under the `[testenv:check_...]` sections, dropping the `deps` and `skip_install` stanzas. Then it should Just Work.

setup.py
H-Shay: Okay all the requested changes have been made and the checks have passed so I hope this is good!

callahad (approved):
Tested everything and it works as expected locally. Yay!

PR #196, opened 2021-04-01
Update CONTRIBUTING.md

tawandamoyo:
Add instruction to install `libq-dev` on Ubuntu

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,9 @@ python3 -m venv venv
 ./venv/bin/pip install -e '.[dev]'
 ```
 
+####*Note:*
+Sygnal depends on the psycop database adapter for PostgreSQL. For the command above to succeed, you may first need to install the development headers for libpq, for example, by running `sudo apt install libpq-dev` on Ubuntu."
+
 This creates an isolated virtual Python environment ("virtualenv") just for
 use with Sygnal, then installs Sygnal along with its dependencies, and lastly
 installs a handful of useful tools


CONTRIBUTING.md
```diff
@@ -15,6 +15,9 @@ python3 -m venv venv
 ./venv/bin/pip install -e '.[dev]'
 ```
 
+####*Note:*
+Sygnal depends on the psycop database adapter for PostgreSQL. For the command above to succeed, you may first need to install the development headers for libpq, for example, by running `sudo apt install libpq-dev` on Ubuntu."
```
callahad, line 19: Can probably get rid of the '####' prefix since we're not really introducing a new heading here; more of an aside.

Also, it's `psycopg2` not `psycop` 🙂 

I did some testing and it's possible that people will need more than just `libpq-dev`; could you please suggest folks install `libpq-dev python3-dev build-essential`?

callahad (changes requested):
Thank you for the contribution! Looks mostly good, though I did have one comment inline and two small requests here. Please:

- Add a [changelog file](https://github.com/matrix-org/sygnal/blob/eb12b62243e40970658d2ac5e3668578abd694c6/CONTRIBUTING.md#changelog) describing this work.
- [Sign-off](https://github.com/matrix-org/sygnal/blob/eb12b62243e40970658d2ac5e3668578abd694c6/CONTRIBUTING.md#sign-off) on your changes so we know we're allowed to use them.

Thank you again 👍

callahad:
Closing; this is being replaced by #197

PR #197, opened 2021-04-01
Update CONTRIBUTING

tawandamoyo:


CONTRIBUTING.md
```diff
@@ -15,6 +15,12 @@ python3 -m venv venv
 ./venv/bin/pip install -e '.[dev]'
 ```
 
+*Note:*
+Sygnal depends on the `psycopg2` database adapter for PostgreSQL. It is recommended that you install `libq-dev python3-dev build-essential` as root, i.e
+
+```sudo apt install libpq-dev python3-dev build-essential```
+
+
```
richvdh, line 23: A few requests:

* could you wrap the text at about 80 characters, in the same way as the rest of the document? Doing so makes it easier to read the raw markdown, and makes it easier to make changes in future.
* For the ``` styling to work properly, you need to use separate lines - see the other examples.
* You need to do this *before* running `pip install`, so I think it would make sense to put it further up in the text.
* This command only works on Debian/Ubuntu, so you might say something like "... which requires the libpq headers and libraries. For example, on Debian/Ubuntu, these can be installed with ..."

richvdh (changes requested):
Thank you for the contribution! I've made a few requests below: please could you make these changes (remember that you can update a pull-request by pushing more updates to your branch).

For our sign-off, we ask that people use their full name. Is `tawanda` your full name, or would you normally use a family name as well? If so, please could you sign any future commits using that.

CONTRIBUTING.md
tawandamoyo: Hi, thanks for the suggestions. On it

CONTRIBUTING.md
```diff
@@ -6,6 +6,13 @@ this case, the [Apache Software License v2](LICENSE).
 
 ## Preparing your development environment
 
+Sygnal depends on the `psycopg2` database adapter for PostgreSQL. 
+It is recommended that you install `libq-dev python3-dev build-essential`. For example on Debian/Ubuntu these can be installed with:
```
callahad, line 10: Thank you for the update -- I think this still needs to be wrapped at 80 characters (your editor should have a setting somewhere that does this automatically)

Otherwise, I'd suggest tweaking the wording very slightly to be a little more precise:

```diff
- It is recommended that you install `libq-dev python3-dev build-essential`.
+ You may need to install development headers for Python and libpq.
```

With those resolved we should be good to go :)

callahad (changes requested):
Looking good! One small change requested.

CONTRIBUTING.md
```diff
@@ -6,6 +6,13 @@ this case, the [Apache Software License v2](LICENSE).
 
 ## Preparing your development environment
 
+Sygnal depends on the `psycopg2` database adapter for PostgreSQL. 
+You may need to install development headers for Python and libpq. For example on Debian/Ubuntu these can be installed with:
```
richvdh, line 10: please could you wrap this line at 80 characters?

richvdh (dismissed):

CONTRIBUTING.md
tawandamoyo: sorry, looked like I did last time. Let me check again and push.

.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
+    "editor.wordWrap": "wordWrapColumn",
+    "[markdown]": {
+
+        "editor.wordWrap": "wordWrapColumn",
+        "editor.wordWrapColumn": 80,
+        "editor.quickSuggestions": false
+    }
+}
```
anoadragon453, line 9: While a good idea, we don't usually include editor-specific files in our repository.

anoadragon453 (commented):
Hello! Thanks for line-wrapping the markdown. Just one small request below and then I think this is good to go.

anoadragon453:
Also, heads up that after making requested changes to your PRs, you can hit the little request review button:
image
(the little circle arrows)

to flag that reviewer to take another look! Otherwise it may take us longer to notice that requested changes have been made :)

.vscode/settings.json
tawandamoyo: Hi, thanks so much. How can I roll that back? Is it possible?

anoadragon453: Yep! If you're using Git with the command line, you can run the following from the repository root to remove the file from git *without* removing it from your filesystem:

```sh
git rm --cached .vscode/settings.json
```

Then just `git add -u && git commit` to make a new commit without the file included :slightly_smiling_face:

anoadragon453: I've gone ahead and done this in 1a54190.

anoadragon453 (approved):
This lgtm!

PR #210, opened 2021-04-04
Ported Buildkite checks to Github actions 

H-Shay:
I ported the first of the CI/CL checks pipeline to Github Actions. I just want to make sure this port was correct before I ported the rest of the checks. Let me know if it is and I will update my pull request with the rest of the checks completed. Thank you! 
Signed-off-by: Hillery Shay shaysquared@gmail.com


clokep:
@H-Shay It looks like this also includes the commits from #194, you probably want to merge the master branch into this branch and then push again!

H-Shay:
Alright I pulled down the upstream repository and merged with my fork and pushed again. Hopefully that's better!

clokep:
> Alright I pulled down the upstream repository and merged with my fork and pushed again. Hopefully that's better!

Looks good now! Thanks!

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,63 @@
+name: CI-Checks
+on: 
+  push:
+    branches-ignore:
+      -master
+      -release-*
```
anoadragon453, line 6: We probably only want to run towncrier on branches that are part of a pull request (as towncrier requires specifying a PR #). This can be achieved in GitHub Actions by specifying:

```yaml
on:
  pull_request:
```

We may want to run other CI tasks just when pushing branches to the repo, even if they aren't part of a visible PR yet though. For that reason, it might be best to split the changelog check into one GitHub actions pipeline, and have the other checks in a separate pipeline file, with a broader `on` definition.

changelog.d/210.misc
```diff
@@ -0,0 +1 @@
+Ported CI/Cl checks to Github Actions
```
anoadragon453, line 1:
```suggestion
Ported CI checks to Github Actions.
```

I'm not sure if "Cl" (lowercase L) is that common a term, and "CI" should encompass changelog checks as well.

tox.ini
```diff
@@ -12,7 +12,7 @@ envlist = py, check_codestyle, check_types
 # creates a symlink to the project directory instead of unpacking the sdist.
 usedevelop=true
 
-commands = trial tests
+commands = trial --coverage tests
```
anoadragon453, line 15: This does match up with other projects (i.e. Synapse), however paradoxically we don't actually make use of test coverage stats at the moment. We'd like to, but the output that `trial` creates isn't too useful in its current form. If we enable test coverage in other places such as Sygnal, we'd like to only do so if the output is meaningful.

So for now we'd probably hold back on this change.

changelog.d/210.misc
anoadragon453, line 1: Additionally, changelog entries should take the imperative form, i.e "Port CI checks to Github Actions." See the [existing changelog](https://github.com/matrix-org/sygnal/blob/master/CHANGELOG.md) for examples.

anoadragon453 (commented):
Thanks for your work; this is a great start! I've left a few comments below.

tox.ini
H-Shay: Yes my bad I did not mean for that change to be added to this PR.

.github/workflows/pipeline.yml
H-Shay: Thanks for the helpful feedback! I updated the PR with this change, there are now two Github Actions file, one with the changelog check and one with the other tests.

changelog.d/210.misc
H-Shay: Sorted!

H-Shay: Cool thanks for letting me know! I've updated the changelog file to reflect this.

.github/workflows/changelog_check.yml
```diff
@@ -0,0 +1,16 @@
+Name: Changelog-Checks
+on: 
+  pull_request:
+    branches-ignore:
+      -master
+      -release-*
```
anoadragon453, line 6: I think we can likely exclude the `branches-ignore` key, as we probably won't put up a PR for a branch with the name `master` or `release-*` against any other branch. Leaving just the empty `pull_request:` key should just run the changelog test against any and all pull requests, which is already an improvement over [the buildkite config](https://github.com/matrix-org/pipelines/blob/e1ae0fb309e88345fb4903d5c4c72b19a47e62a6/sygnal/pipeline.yml#L10).

.github/workflows/pipeline.yml
```diff
@@ -1,55 +1,49 @@
...
+    steps: 
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
+      - run: python -m pip install -e .
+      - run: trial tests
+  
+  # run-unit-tests-(on :postgres:):
+  #   needs: [run-unit-tests-sqlite]
```
anoadragon453, line 36: Having the postgres test `need`ing the sqlite test will cause them to run sequentially, whereas ideally they would run in parallel. I'm not sure if there's a better way to structure this other than adding `needs: [check-code-style, check-types-mypy]` to each job.

.github/workflows/pipeline.yml
```diff
@@ -1,55 +1,49 @@
...
+    runs-on: ubuntu-latest
+    steps: 
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
+      - run: python -m pip install -e .
+      - run: trial tests
+  
+  # run-unit-tests-(on :postgres:):
```
anoadragon453, line 35: I'm not sure if Github Actions supports custom emoji, but `:postgres:` on buildkite would translate into a picture of the postgres logo, as can be seen here: https://buildkite.com/matrix-dot-org/synapse/builds/16518#4b64142c-9740-4104-9009-7df4230abd54

In case you were wondering :blush:

anoadragon453, line 35: Also, [GA job services](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservices) look like they may be a good way to spin up a running Postgres instance and pass the necessary env vars.

anoadragon453 (commented):
I'm assuming you wanted another review of this PR since you answered all my questions, but apologies if you were still working on this round! And if you were waiting for review, note that hitting the request review button is the best way to communicate that :slightly_smiling_face:

That said, this is starting to really shape up! I'll popped some hopefully help notes below.

.github/workflows/pipeline.yml
callahad: Alas, GHA does not support custom emoji names for jobs ☹️

.github/workflows/pipeline.yml
callahad: > I'm not sure if there's a better way to structure this other than adding needs: `[check-code-style, check-types-mypy]` to each job.

There isn't, unfortunately. In Synapse I made a dummy task called `linting-done` like:

```yaml
  # Dummy step to gate other tests on without repeating the whole list
  linting-done:
    needs: [...]
    runs-on: ubuntu-latest
    steps:
      - run: "true"
```

And then all the subsequent jobs could just `needs: linting-done`

But that's almost certainly overkill for Sygnal's test matrix

.github/workflows/changelog_check.yml
H-Shay: Thanks for the notes-everything should be fixed now! I also was not aware of the request review button, thanks so much for telling me about that.

.github/workflows/pipeline.yml
```diff
@@ -32,26 +32,37 @@ jobs:
...
+      - run: python -m pip install -e .
+      - run: trial tests
+        env:
+          TEST_USE_POSTGRES: "yes"
+          TEST_POSTGRES_USER: "postgres"
+          TEST_POSTGRES_PASSWORD: "postgres"
+          TEST_POSTGRES_HOST: "localhost"
+      
+      
+
```
anoadragon453, line 67:
```suggestion
```

Just getting rid of some unneeded newlines.

.github/workflows/changelog_check.yml
```diff
@@ -1,9 +1,6 @@
 Name: Changelog-Checks
```
anoadragon453, line 1:
```suggestion
name: Changelog
```

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,68 @@
+name: CI-Checks
```
anoadragon453, line 1:
```suggestion
name: Linting and Tests
```

Just to be a bit more descriptive. I'm *pretty* these can have spaces in them.

.github/workflows/changelog_check.yml
```diff
@@ -0,0 +1,13 @@
...
+on: [pull_request]
+
+jobs:
+  check-newsfile:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
```
anoadragon453, line 11:
```suggestion
          python-version: "3.7"
```

And in the other pipeline, so that we're consistent with our strings.

anoadragon453 (changes requested):
Looks good to me other than a few small things!

anoadragon453 (approved):
Heads up that if you've resolved an issue as part of a pull request comment thread, resolving the thread once discussion has finished and the suggestion has been implemented (or rejected) helps shrink the total vertical space of the PR.

Anyways, this PR looks ready to go :tada:  Thank you for your work on it! :slightly_smiling_face:

anoadragon453:
Oh, actually, as a very pedantic last step, can I get you to sign off on your changes just in a comment stating your ownership of this code change? Details for doing so are here: https://github.com/matrix-org/sygnal/blob/master/CONTRIBUTING.md#sign-off

H-Shay:
Cool! I updated the original PR comment to include a sign-off statement. Thanks for all your helpful feedback!

Issue #186, opened 2021-03-26
Web push improvements


bwindels:
- [ ] Set `Urgency` header based on `prio` field in notification:
   - "high" or absent => "normal"
   - "low" => "low"
- [ ] Don't reject push key on every 4xx code
- [ ] Trim message body so we don't exceed 4096 characters
- [ ] Log when returned `TTL` header is lower than request

bwindels:
https://autopush.readthedocs.io/en/latest/http.html#error-codes

H-Shay:
Looks like these tasks are done, can we close this issue?

matrixbot:
For your information, this issue has been copied over to the Element fork of `sygnal`: <https://github.com/element-hq/sygnal/issues/186>

PR #213, opened 2021-04-07
WebPush support for urgency and coalescing

bwindels:
Related to https://github.com/matrix-org/sygnal/issues/186, based off #212 :
 - support dropping earlier notifications in the same room using the [`Topic` header](https://tools.ietf.org/html/draft-ietf-webpush-protocol-12#section-5.4)
 - set the `Urgency` header based on the notification prio field.

### DIFF
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -270,7 +270,7 @@ forced to show a notification to push messages that clear the unread count.
 ##### only_last_per_room
 
 You can opt in to only receive the last notification per room by setting
-`only_last_per_room: true` in the push data. Note that if a first notification
+`only_last_per_room: true` in the push data. Note that if the first notification
 can be delivered before the second one is sent, you will still get both;
 it only has an effect when notifications are queued up on the gateway.
 
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -257,6 +257,10 @@ so it can be used to pass identifiers specific to your client
 
 As of the time of writing, all webpush-supporting browsers require you to set `userVisibleOnly: true` when calling (`pushManager.subscribe`)[https://developer.mozilla.org/en-US/docs/Web/API/PushManager/subscribe], to (prevent abusing webpush to track users)[https://goo.gl/yqv4Q4] without their knowledge. With this (mandatory) flag, the browser will show a "site has been updated in the background" notification if no notifications are visible after your service worker processes a `push` event. This can easily happen when sygnal sends a push message to clear the unread count, which is not specific to an event. With `events_only: true` in the pusher data, sygnal won't forward any push message without a event id. This prevents your service worker being forced to show a notification to push messages that clear the unread count.
 
+##### only_last_per_room
+
+You can opt in to only receive the last notification per room by setting `only_last_per_room: true` in the push data. Note that if a first notification can be delivered before the second one is sent, you will still get both; it only has effect for when notifications are queued up on the gateway.
+
 ##### Multiple pushers on one origin
 
 Also note that because you can only have one push subscription per service worker,
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -272,7 +272,7 @@ forced to show a notification to push messages that clear the unread count.
 You can opt in to only receive the last notification per room by setting
 `only_last_per_room: true` in the push data. Note that if a first notification
 can be delivered before the second one is sent, you will still get both;
-it only has effect for when notifications are queued up on the gateway.
+it only has an effect when notifications are queued up on the gateway.
 
 ##### Multiple pushers on one origin
 
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -255,11 +255,24 @@ so it can be used to pass identifiers specific to your client
 
 ##### events_only
 
-As of the time of writing, all webpush-supporting browsers require you to set `userVisibleOnly: true` when calling (`pushManager.subscribe`)[https://developer.mozilla.org/en-US/docs/Web/API/PushManager/subscribe], to (prevent abusing webpush to track users)[https://goo.gl/yqv4Q4] without their knowledge. With this (mandatory) flag, the browser will show a "site has been updated in the background" notification if no notifications are visible after your service worker processes a `push` event. This can easily happen when sygnal sends a push message to clear the unread count, which is not specific to an event. With `events_only: true` in the pusher data, sygnal won't forward any push message without a event id. This prevents your service worker being forced to show a notification to push messages that clear the unread count.
+As of the time of writing, all webpush-supporting browsers require you to set 
+`userVisibleOnly: true` when calling (`pushManager.subscribe`)
+[https://developer.mozilla.org/en-US/docs/Web/API/PushManager/subscribe], to 
+(prevent abusing webpush to track users)[https://goo.gl/yqv4Q4] without their 
+knowledge. With this (mandatory) flag, the browser will show a "site has been 
+updated in the background" notification if no notifications are visible after
+your service worker processes a `push` event. This can easily happen when sygnal
+sends a push message to clear the unread count, which is not specific
+to an event. With `events_only: true` in the pusher data, sygnal won't forward
+any push message without a event id. This prevents your service worker being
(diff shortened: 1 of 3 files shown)


docs/applications.md
```diff
@@ -257,6 +257,10 @@ so it can be used to pass identifiers specific to your client
 
 As of the time of writing, all webpush-supporting browsers require you to set `userVisibleOnly: true` when calling (`pushManager.subscribe`)[https://developer.mozilla.org/en-US/docs/Web/API/PushManager/subscribe], to (prevent abusing webpush to track users)[https://goo.gl/yqv4Q4] without their knowledge. With this (mandatory) flag, the browser will show a "site has been updated in the background" notification if no notifications are visible after your service worker processes a `push` event. This can easily happen when sygnal sends a push message to clear the unread count, which is not specific to an event. With `events_only: true` in the pusher data, sygnal won't forward any push message without a event id. This prevents your service worker being forced to show a notification to push messages that clear the unread count.
 
+##### only_last_per_room
+
+You can opt in to only receive the last notification per room by setting `only_last_per_room: true` in the push data. Note that if a first notification can be delivered before the second one is sent, you will still get both; it only has effect for when notifications are queued up on the gateway.
```
richvdh, line 262: wrap at 80 chars please

sygnal/webpushpushkin.py
```diff
@@ -173,6 +175,17 @@ async def _dispatch_notification_unlimited(self, n, device, context):
         payload = WebpushPushkin._build_payload(n, device)
         data = json.dumps(payload)
 
+        # web push only supports normal and low priority, so assume normal if absent
+        low_priority = n.prio == "low"
+        # allow dropping earlier notifications in the same room if requested
+        topic = None
+        if n.room_id and device.data.get("only_last_per_room") is True:
```
richvdh, line 182:
```suggestion
        if n.room_id and device.data.get("only_last_per_room") == True:
```

sygnal/webpushpushkin.py
```diff
@@ -336,25 +348,15 @@ def post(self, endpoint, data, headers, timeout):
...
 
-class HttpResponseWrapper:
+class HttpDelayedRequest:
     """
+    Captures the values received from pywebpush for the endpoint request.
+    The request isn't immediately executed, to allow adding headers
+    not supported by pywebpush, like Topic and Urgency.
+
     Provide a response object that matches the API expected from pywebpush.
     pywebpush expects a synchronous API, while we use an asynchronous API.
```
richvdh, line 361: this looks outdated?

sygnal/webpushpushkin.py
```diff
@@ -375,5 +375,26 @@ class HttpResponseWrapper:
     status_code = 200
     text = None
```
richvdh, line 376: are these still used?

sygnal/webpushpushkin.py
```diff
@@ -186,15 +199,17 @@ async def _dispatch_notification_unlimited(self, n, device, context):
...
                         data=data,
                         ttl=self.ttl,
                         vapid_private_key=self.vapid_private_key,
                         vapid_claims=vapid_claims,
-                        requests_session=self.http_agent_wrapper,
+                        requests_session=self.http_request_factory,
+                    )
+                    response = await request.execute(
+                        self.http_agent, low_priority, topic
                     )
```
richvdh, line 212: where does the deduplication actually take place, out of interest?

sygnal/webpushpushkin.py
bwindels: This still serves as something that looks like a response to pywebpush, while actually containing the information to perform the request ourselves once pywebpush finishes. I agree the naming is not ideal, but couldn't come up with something better. Ideas?

sygnal/webpushpushkin.py
bwindels: Yes, as mentioned above.

sygnal/webpushpushkin.py
bwindels: We don't do this, it happens in the push gateway. If you have two messages with the same topic header, the second will replace the first (if the first hasn't been delivered yet).

sygnal/webpushpushkin.py
bwindels: Actually, CI tells me it prefers the former way:
```
sygnal/webpushpushkin.py:182:64: E712 comparison to True should be 'if cond is True:' or 'if cond:'
```
So I'll revert?

richvdh: huh. my bad, sorry.

sygnal/webpushpushkin.py
richvdh: ah, so there is a separate "push gateway" after sygnal (which I tend to think of as a, y'know, push gateway)?

docs/applications.md
```diff
@@ -255,11 +255,24 @@ so it can be used to pass identifiers specific to your client
...
+sends a push message to clear the unread count, which is not specific
+to an event. With `events_only: true` in the pusher data, sygnal won't forward
+any push message without a event id. This prevents your service worker being
+forced to show a notification to push messages that clear the unread count.
 
 ##### only_last_per_room
 
-You can opt in to only receive the last notification per room by setting `only_last_per_room: true` in the push data. Note that if a first notification can be delivered before the second one is sent, you will still get both; it only has effect for when notifications are queued up on the gateway.
+You can opt in to only receive the last notification per room by setting
+`only_last_per_room: true` in the push data. Note that if a first notification
```
richvdh, line 273:
```suggestion
`only_last_per_room: true` in the push data. Note that if the first notification
```

docs/applications.md
```diff
@@ -255,11 +255,24 @@ so it can be used to pass identifiers specific to your client
...
+can be delivered before the second one is sent, you will still get both;
+it only has effect for when notifications are queued up on the gateway.
```
richvdh, line 275:
```suggestion
it only has an effect when notifications are queued up on the gateway.
```

sygnal/webpushpushkin.py
richvdh: right, I see.

sygnal/webpushpushkin.py
```diff
@@ -336,25 +348,15 @@ def post(self, endpoint, data, headers, timeout):
...
 
 
-class HttpResponseWrapper:
+class HttpDelayedRequest:
     """
+    Captures the values received from pywebpush for the endpoint request.
+    The request isn't immediately executed, to allow adding headers
+    not supported by pywebpush, like Topic and Urgency.
+
     Provide a response object that matches the API expected from pywebpush.
```
richvdh, line 360:
```suggestion
    Also provides the interface that pywebpush expects from a response object.
```

sygnal/webpushpushkin.py
```diff
@@ -363,8 +365,6 @@ class HttpResponseWrapper:
     in the background.
```
richvdh, line 365: it now doesn't really happen "in the background", afaict? it's more that it is deferred until later.

How about:

>    To keep pywebpush happy we present it with some hardcoded values that
>    make its assertions pass even though the HTTP request has not yet been
>    made.

richvdh (commented):
lgtm generally, a few nits

sygnal/webpushpushkin.py
bwindels: I mean, webpush pipes notification to a push provider (or gateway, hard to name things correctly with so many parts to it) like FCM that will actually deliver it to the browser application, similarly to how the apns and gcm pushkins forward it to the push providers of those platforms. Every browser that supports web push needs to have a server like this that application servers can post push messages to. It's the browser-specific push provider that will drop the first message if it hasn't been deliver yet.

sygnal/webpushpushkin.py
bwindels: Indeed, I forgot to update this as the async request was fired immediately before but now is deferred to the execute method :+1: Have adopted your suggestion.

richvdh (approved):
lgtm, sorry for the slow review!

PR #215, opened 2021-04-12
Connection Timeout Error

kibablu:
Add connection timeout error code on `CONTRIBUTING.mb`, the connection timeout
error is caused by poor internet connectivity as `pip` is set to default 15 sec.

When running this command with slow internet then `pip` throws a connection time error as `pip` is set to the default of 15 sec to download any dependency
```bash
python3 -m venv venv
```


CONTRIBUTING.md
```diff
@@ -19,6 +19,16 @@ This creates an isolated virtual Python environment ("virtualenv") just for
 use with Sygnal, then installs Sygnal along with its dependencies, and lastly
 installs a handful of useful tools
 
+If you get `ConnectTimeoutError`, this is caused by slow internet whereby `pip` has a 
+default time out of _15 sec_. Run the following command by specifying timeout to be more
+than 15 sec.
+
+```bash
+python3 -m venv venv
+./venv/bin/pip timeout==1000 install -e '.[dev]'
```
callahad, line 28: I think `timeout==1000` should instead be `--timeout 1000`, right?

And maybe lower the value to 60 or 120, at your discretion.

If pip doesn't receive anything in over a minute it's probably a sign that something is wrong elsewhere. Want to be mindful of striking a balance between allowing enough time and detecting errors. E.g., if PyPI is broken, it's probably not right for people to wait nearly 17 minutes (timeout 1000) for things to fail.

changelog.d/215.doc
```diff
@@ -0,0 +1 @@
+Add connection timeout error code  #215. Contributed by Omar Mohamed
```
callahad, line 1: For some reason we really want these changelog entries to end with a . or ! mark -- please add one :)

CONTRIBUTING.md
```diff
@@ -19,6 +19,16 @@ This creates an isolated virtual Python environment ("virtualenv") just for
 use with Sygnal, then installs Sygnal along with its dependencies, and lastly
 installs a handful of useful tools
 
+If you get `ConnectTimeoutError`, this is caused by slow internet whereby `pip` has a 
+default time out of _15 sec_. Run the following command by specifying timeout to be more
+than 15 sec.
```
callahad, line 24: Instead of repeating the code block, maybe just add the timeout note in line here?

E.g., instead of saying "Run the following command by specifying timeout to be more than 15 sec." you could say "You can specify a larger timeout by passing `--timeout 120` to the `pip install` command above."

callahad (changes requested):
Hey, thank you so much! I've requested a few changes, after which this should be good to go

CONTRIBUTING.md
kibablu: > If pip doesn't receive anything in over a minute it's probably a sign that something is wrong elsewhere. Want to be mindful of striking a balance between allowing enough time and detecting errors. E.g., if PyPI is broken, it's probably not right for people to wait nearly 17 minutes (timeout 1000) for things to fail.

Thank you, I will have `timeout` to 60 sec

> I think timeout==1000 should instead be --timeout 1000, right?

I will update `timeout` without the `equal sign` just seen that on `pip --help` command

CONTRIBUTING.md
kibablu: got it, Thank you

callahad (approved):
Great! Thank you so much 💯

changelog.d/214.misc
```diff
@@ -1 +1 @@
-Upgrade development dependencies fixes #199. Contributed by Omar Mohamed
+Upgrade development dependencies fixes #199. Contributed by Omar Mohamed.
```
callahad, line 1: Oops! I noticed one final thing right before I hit merge 😆

Could you please rename this file to `215.misc` instead of `214.misc`, and maybe change the description to more closely match (e.g., "Add note about pip timeouts. Contributed by Omar Mohamed.")

callahad (changes requested):

changelog.d/214.misc
kibablu: On it :smiley:

callahad: 215.doc here would be fine :)

kibablu: I have made changes on the `215.doc` changelog file with a more descriptive message on `pip timeout` :smiley:

callahad (approved):
Works for me -- thank you so much 🎉

callahad:
I manually merged this in 7a3d6f8f47229ce8d03499b186ca8f1906797843 with a few small tweaks -- thank you again 😄

kibablu:
> I manually merged this in [7a3d6f8](https://github.com/matrix-org/sygnal/commit/7a3d6f8f47229ce8d03499b186ca8f1906797843) with a few small tweaks -- thank you again

Thank you so much :smiley:

PR #216, opened 2021-04-12
Be stricter about JSON that is accepted by Sygnal

richvdh:
This disables the JSON extensions which Python supports by default (parsing of
`Infinity` / `-Infinity` and `NaN`). These shouldn't be accepted since they're
not technically valid JSON and other languages won't be able to interpret it
properly.

### DIFF
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -32,7 +32,7 @@ from sygnal.exceptions import (
 )
 from sygnal.helper.context_factory import ClientTLSOptionsFactory
 from sygnal.helper.proxy.proxyagent_twisted import ProxyAgent
-from sygnal.utils import NotificationLoggerAdapter, json_decoder, twisted_sleep
+from sygnal.utils import NotificationLoggerAdapter, safe_json_loads, twisted_sleep
 
 from .exceptions import PushkinSetupException
 from .notifications import ConcurrencyLimitedPushkin
@@ -250,7 +250,7 @@ class GcmPushkin(ConcurrencyLimitedPushkin):
             return pushkeys, []
         elif 200 <= response.code < 300:
             try:
-                resp_object = json_decoder.decode(response_text)
+                resp_object = safe_json_loads(response_text)
             except ValueError:
                 raise NotificationDispatchException("Invalid JSON response from GCM.")
             if "results" not in resp_object:
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -32,7 +32,7 @@ from sygnal.exceptions import (
 )
 from sygnal.helper.context_factory import ClientTLSOptionsFactory
 from sygnal.helper.proxy.proxyagent_twisted import ProxyAgent
-from sygnal.utils import NotificationLoggerAdapter, safe_json_loads, twisted_sleep
+from sygnal.utils import NotificationLoggerAdapter, json_decoder, twisted_sleep
 
 from .exceptions import PushkinSetupException
 from .notifications import ConcurrencyLimitedPushkin
@@ -250,7 +250,7 @@ class GcmPushkin(ConcurrencyLimitedPushkin):
             return pushkeys, []
         elif 200 <= response.code < 300:
             try:
-                resp_object = safe_json_loads(response_text)
+                resp_object = json_decoder.decode(response_text)
             except ValueError:
                 raise NotificationDispatchException("Invalid JSON response from GCM.")
             if "results" not in resp_object:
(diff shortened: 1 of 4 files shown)


sygnal/gcmpushkin.py
```diff
@@ -251,8 +251,8 @@ async def _request_dispatch(self, n, log, body, headers, pushkeys, span):
             return pushkeys, []
         elif 200 <= response.code < 300:
             try:
-                resp_object = json.loads(response_text)
-            except JSONDecodeError:
```
callahad, line 255: This is causing flake8 to fail, since the import is still present on line 21. (`F401 'json.JSONDecodeError' imported but unused`)

callahad (changes requested):

sygnal/gcmpushkin.py
callahad: Fixed and pushed to your branch

callahad:
I rewrote the json decoder to use `functools.partial` to bind `json.loads` instead of instantiating a new `JSONDecoder`. Tests pass now, and it behaves as expected:

```python
>>> safe_json_loads('{"hello": "world"}')
{'hello': 'world'}

>>> safe_json_loads('{"hello": Infinity}')
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/json/__init__.py", line 361, in loads
    return cls(**kw).decode(s)
  File "/usr/local/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/lib/python3.7/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
  File "<stdin>", line 3, in _reject_invalid_json
ValueError: Invalid JSON value: 'Infinity'
```

callahad (approved):

sygnal/utils.py
```diff
@@ -62,3 +64,12 @@ def glob_to_regex(glob):
...
     return re.compile(r"\A" + res + r"\Z", re.IGNORECASE)
+
+
+def _reject_invalid_json(val):
+    """Do not allow Infinity, -Infinity, or NaN values in JSON."""
+    raise ValueError(f"Invalid JSON value: {val!r}")
+
+
+# a custom JSON decoder which will reject Python extensions to JSON.
+safe_json_loads = partial(json.loads, parse_constant=_reject_invalid_json)
```
richvdh, line 75: given that `json.loads` instantiates a `JSONDecoder`, why this is an improvement?

(Also: we do the local `json_decoder` thing in Sydent and Synapse, and there is value in being consistent)

callahad: > why this is an improvement?

Because it works 🙂  cf [failures in CI](https://buildkite.com/matrix-dot-org/sygnal/builds/569#f5a3ca95-3919-4f90-af63-e2b05fcc29f0/53-453) with the previous solution.

Keeping aligned with `loads` fits my brain better, but I don't have strong feelings at all so long as the tests pass.

callahad: Upon further reflection, I think I have a very slight preference for the `partial` solution, since it is literally a drop-in replacement for `json.loads`, while `JSONDecoder.decode`, empirically, is not.

The actual difference is that `json.loads` accepts `str | bytes` and [handles decoding](https://github.com/python/cpython/blob/3.6/Lib/json/__init__.py#L341-L349) while `JSONDecoder.decode` only accepts `str`.

richvdh: ok, I'd rather avoid constructing a new JSONDecoder every call if possible. I've tweaked this to use a static `json_decoder` whilst still (hopefully) passing CI.

callahad: > I'd rather avoid constructing a new JSONDecoder every call if possible

Seems like a premature optimization, but tests pass, so if you're happy I'm happy :)

callahad (approved):

sygnal/utils.py
richvdh: Sorry, our comments have been crossing.

Personally, I would rather we were type-strict and were certain about whether we had `bytes` or `str`s, rather than using interfaces which handle both. (Incidentally: I now realise the reason this didn't bite us on Synapse or Sydent is that both of those support python 3.5, where `json.loads` does *not* accept `bytes`, so we had already done the explicit `.decode()` calls).

So, to summarise, I see the following in favour of this approach:

* consistency with Synapse and Sydent
* type-strictness
* avoids constructing a JSONDecoder on each call (yes, a tiny optimisation, but stuff like this adds up, especially on Synapse)

So, thank you for humouring me :)

Issue #200, opened 2021-04-02
Measure code coverage


callahad:
We should find a way to measure the code coverage of our unit tests (e.g., by running our tests under [coverage.py](https://coverage.readthedocs.io/en/coverage-5.5/) or using Twisted Trial's built in `--coverage` flag)

H-Shay:
I updated tox.ini in my fork to use `commands = trial --coverage tests` to use the built-in `--coverage` flag. It ran all the tests successfully with this change. If this is all that needs to be done let me know and I will open a PR with the change.

anoadragon453:
@callahad I'm a bit concerned that the coverage for `trial`in Synapse today isn't really used by anyone on the team, and at least I find it to be more noise than anything else. We talked about having a different interface in the past that allowed tracking coverage changes over time, as well as allowing to peer into a file and see the coverage line-for-line, which may be more useful that just seeing a snapshot of things at the file level.

It would be nice if they were placed elsewhere other than the end of the unit test output, as I often find myself needing to scroll up past it to get to the more immediate need: the failing test output.

This probably isn't the right issue to discuss making the output more useful, but I wonder if we want to hold back on adding coverage output to other projects (i.e Sygnal) until we've got a meaningful setup for it on Synapse.

callahad:
> We talked about having a different interface in the past that allowed tracking coverage changes over time, as well as allowing to peer into a file and see the coverage line-for-line

This is exactly where I want to end up. It seems like Trial's built-in reporting doesn't do much for us in that regard, but [coverage.py](https://coverage.readthedocs.io/en/coverage-5.5/) has a nice `coverage html` command which spits out a pretty enough report:

![Screen Shot 2021-04-14 at 11 10 34](https://user-images.githubusercontent.com/24193/114694320-4fe03d00-9d12-11eb-9daa-d44fabaf33ff.png)

which you can then click around in to see line-by-line information:

![Screen Shot 2021-04-14 at 11 13 17](https://user-images.githubusercontent.com/24193/114694445-700ffc00-9d12-11eb-8515-0268069b804e.png)

> I wonder if we want to hold back on adding coverage output to other projects (i.e Sygnal) until we've got a meaningful setup for it on Synapse.

I think it might be nice to figure things out in a lower stakes repo (Sygnal) before pushing that rock all the way up the hill that is Synapse 🙂

callahad:
[...]

PR #217, opened 2021-04-15
setup coverage.py in tox env

H-Shay:
Hopefully this satisfies the first part of https://github.com/matrix-org/sygnal/issues/200#issuecomment-819410266. If so, I'll move on to adding reports to Github actions, etc. Thanks!


H-Shay:
Noting that the towncrier github actions check is failing. Looks like an error with a config file somewhere (git produced output while failing:
fatal: ambiguous argument 'origin/master...': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]')
I didn't experience this error when setting up the github actions so I'm a little perplexed, but I will look deeper into it tomorrow.

anoadragon453:
The command that `towncrier` is running is `git diff --name-only origin/master...`, and failing to figure out what `origin/master` is. There's a lot of possible reasons for this. `checkout@v2` might be specifying your fork as `origin`, and may only be checkout out your `add-coverage` branch but nothing else. It may be possible to fix this by just doing a `git fetch origin` before running `towncrier`.

I'd probably play around in this PR (or another) and try to get some information out of GitHub Actions in terms of how it's set it's remotes and branches up.

callahad:
Looks good!

- Let's add `.coverage*` and `htmlcov/` to the `.gitignore` file and remove them from this pull request; no need to ship the reports with our source.
- It would probably be good to run `coverage report --sort=cover` _and_ `coverage html` so that there's something in the console as well as the html if you want to delve deeper.
- ...and maybe only run the report(s) if the initial `-m twisted.trial` succeeds?

callahad:
As to the towncrier bit... by default GitHub Actions does a shallow clone of the repo which has only one commit: the result of the proposed merge. So there's no actual history there to compare against. Solving that properly is a bit of a rabbit hole, but I _think_ setting:

```
- uses: actions/checkout@v2
  with:
    fetch-depth: 0
    ref: ${{ github.event.pull_request.head.sha }}
```

would get you a full history, check out the actual branch head (not the proposed merge) and allow the check to work.

H-Shay:
Ok cool I will open a different PR for the towncrier issue and see if I can't get it working using the insight you both offered. 

I've updated the PR to add the other changes related to coverage. However I have one question:

In terms only running the coverage reports if the `-m twisted.trial` succeeds, I am wondering if tox doesn't do this automatically? Looking at the docs, under **commands** it states that "Commands will execute one by one in sequential fashion until one of them fails (their exit code is non-zero) or all of them succeed." My understanding of this is that tox will exit if the `twisted.trial` fails. I am unfamiliar with tox so please correct me if I'm wrong!

callahad:
> "Commands will execute one by one in sequential fashion until one of them fails (their exit code is non-zero) or all of them succeed."

Oh, cool! You're probably totally good there, then :)

H-Shay:
Great then this should be all set to go!

richvdh:
@callahad I think you have most context here so if you have the cycles to pick up the review that would be appreciated!

callahad (changes requested):
Great, thanks! Could you please:

1. Merge `origin/master` (or rebase onto `origin/master`) so we can get working towncrier checks
2. Add `coverage~=5.5` as a dependency somewhere to ensure that a clean`tox` virtual environment has access to it? (e.g., it could be a dev dependency in setup.py, or a one-off `deps` entry in the `tox.ini` testenv.

H-Shay:
Alright I think that did it. Thank you!

callahad:
We're almost good :) The `tox.ini` still needs to be adjusted to actually install `coverage` in the default testenv. You can just remove the `extras =` stanza from the other environments and put it straight in the base `testenv` and then all three will inherit it and install the `dev` dependencies.

H-Shay:
Thanks for the hand-holding, I am actually learning a lot. Hopefully this is the last update!

callahad (approved):

callahad:
Thank you so much!

callahad:
Oh, hey, could you sign-off on a commit in this branch, please?

You can do a `git commit --signoff --allow-empty` to make an empty commit to that effect :)

PR #219, opened 2021-04-15
fix towncrier fail in github actions

H-Shay:
first attempt fixing towncrier failure in github actions

### DIFF
--- a/.github/workflows/changelog_check.yml
+++ b/.github/workflows/changelog_check.yml
@@ -6,6 +6,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: ${{github.event.pull_request.head.sha}}
       - uses: actions/setup-python@v2
         with:
           python-version: "3.7"


changelog.d/219.bugfix
```diff
@@ -0,0 +1 @@
+fix towncrier failure in Github Actions
```
richvdh, line 1: A couple of things here. 

Firstly: this isn't fixing a bug that a user would see - it's more of a development thing, so it would be better to call this `219.misc`, so that it ends up in the "Internal changes" section of the changelog.

But in this case, since we haven't made a release since #210, it might be more appropriate to give it the same changelog as #210. (Just create a `219.misc` with exactly the same contents as `210.misc`). Then, when we come to do the release, towncrier will merge the two entries together and link to both PRs.

richvdh (changes requested):
thanks for this! The change itself looks good, but would you mind updating the changelog entry?

changelog.d/219.bugfix
H-Shay: Cool thanks for reviewing! I made the requested changes and updated the PR.

richvdh (approved):
looks great. Thanks very much!

PR #221, opened 2021-04-21
Add improved static type checking in setup.py

kibablu:
Add static type checking on setup.py with `Any` as the path-like object is either `str` or `bytes` object representing a path, or an object implementing the os path.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,8 +16,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-from typing import Any
+from typing import Union
+from os import PathLike
+import os.path
 
 from setuptools import find_packages, setup
 
@@ -26,7 +27,7 @@ from setuptools import find_packages, setup
 # Used for the long_description.  It's nice, because now 1) we have a top level
 # README file and 2) it's easier to type in the README file than to put a raw
 # string in below ...
-def read(fname: Any) -> Any:
+def read(fname: Union[str, PathLike[str]]) -> str:
     return open(os.path.join(os.path.dirname(__file__), fname)).read()
 
 
--- a/setup.py
+++ b/setup.py
@@ -16,9 +16,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Union
-from os import PathLike
 import os.path
+from os import PathLike
+from typing import Union
 
 from setuptools import find_packages, setup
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -25,7 +26,7 @@
 # Used for the long_description.  It's nice, because now 1) we have a top level
 # README file and 2) it's easier to type in the README file than to put a raw
 # string in below ...
-def read(fname):
+def read(fname: Any) -> Any:
     return open(os.path.join(os.path.dirname(__file__), fname)).read()
```
callahad, line 30: Unfortunately, declaring `Any` doesn't help us catch invalid values. For example, this function would throw an exception if passed an integer, but mypy wouldn't warn us ahead of time since everything matches `Any`.

What do you think these types should be?

callahad (changes requested):

setup.py
kibablu: I was thinking of `str` but we also have `bytes` objects representing a path, is it possible to combine `str` and `bytes`

callahad:
I'm surprised this is failing.

Let me test if I can get this to blow up locally, and if so, whether adding `from __future__ import annotations` fixes it or if there's something else we need to do.

callahad:
Okay, yep. This is failing for a very obscure reason. Python 3.7 and 3.8 don't understand the `[str]` part of `PathLike[str]`, and so they throw an error.

We can fix that by opting into new behaviors for type annotations. Making the following change should get the tests to pass:

```patch
diff --git a/setup.py b/setup.py
index 88a79f5..2e91d7a 100755
--- a/setup.py
+++ b/setup.py
@@ -16,9 +16,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Union
-from os import PathLike
+from __future__ import annotations
+
 import os.path
+from os import PathLike
+from typing import Union
 
 from setuptools import find_packages, setup
 
```

...and since we're using the `__future__` import, we can also use the new shorthand syntax for Unions:

```patch
diff --git a/setup.py b/setup.py
index 2e91d7a..303c791 100755
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,6 @@ from __future__ import annotations
 
 import os.path
 from os import PathLike
-from typing import Union
 
 from setuptools import find_packages, setup
 
@@ -29,7 +28,7 @@ from setuptools import find_packages, setup
 # Used for the long_description.  It's nice, because now 1) we have a top level
 # README file and 2) it's easier to type in the README file than to put a raw
 # string in below ...
-def read(fname: Union[str, PathLike[str]]) -> str:
+def read(fname: str | PathLike[str]) -> str:
     return open(os.path.join(os.path.dirname(__file__), fname)).read()
 
 
```

kibablu:
Thank you @callahad  :smiley:  :tada: :tada:

callahad (approved):
And the tests pass! Yay! Thank you so much

PR #223, opened 2021-04-29
Improve static type checking

### LINKED ISSUE #201: Improve static type checking
Running `mypy --strict sygnal/ tests/ setup.py` returns "634 errors in 27 files". We should start chipping away at that little by little. :)

For an introduction to mypy, see https://mypy.readthedocs.io/

kibablu:
Add type annotation to `json_encode Function` on `apnstruncate.py` file

### DIFF
--- a/sygnal/apnstruncate.py
+++ b/sygnal/apnstruncate.py
@@ -15,10 +15,10 @@
 # Copied and adapted from
 # https://raw.githubusercontent.com/matrix-org/pushbaby/master/pushbaby/truncate.py
 import json
-from typing import List, Tuple, Union
+from typing import Any, Dict, List, Tuple, Union
 
 
-def json_encode(payload):
+def json_encode(payload: Dict[str, Any]) -> bytes:
     return json.dumps(payload, ensure_ascii=False).encode()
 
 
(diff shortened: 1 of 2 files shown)


sygnal/apnstruncate.py
```diff
@@ -15,10 +15,10 @@
 # Copied and adapted from
 # https://raw.githubusercontent.com/matrix-org/pushbaby/master/pushbaby/truncate.py
 import json
-from typing import List, Tuple, Union
+from typing import Any, Dict, List, Tuple, Union
 
 
-def json_encode(payload):
+def json_encode(payload: Dict[str, Any]) -> bytes:
```
babolivier, line 21: `payload` doesn't have to be a dict, it can be anything that can be translated into json (so pretty much any native type). I think the right type for it would be `Any`.

babolivier (changes requested):
Thanks a lot for your contribution, there's one small change to make but overall it looks good to me :)
I'm also going to edit the PR description so it doesn't accidentally close #201 (since this is only addressing part of the issue).

sygnal/apnstruncate.py
kibablu: Now payload should look like this `(payload: Any) -> bytes` or it should also return `Any `

babolivier (approved):
Looks good to me! Thanks again for your contribution :)

sygnal/apnstruncate.py
babolivier: `bytes` is the right value, thanks for changing the payload type :)

kibablu:
Thank you :)

PR #225, opened 2021-05-24
Improve static type checking

kibablu:
Add type annotation to is_too_long Function on apnstruncate.py file
as part of the #201 issue

### DIFF
--- a/sygnal/apnstruncate.py
+++ b/sygnal/apnstruncate.py
@@ -26,7 +26,7 @@ class BodyTooLongException(Exception):
     pass
 
 
-def is_too_long(payload: Dict[Any, Any], max_length: int = 2048) -> bool:
+def is_too_long(payload: dict, max_length: int = 2048) -> bool:
     """
     Returns True if the given payload dictionary is too long for a push.
     Note that the maximum is now 2kB "In iOS 8 and later" although in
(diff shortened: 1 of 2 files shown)


sygnal/apnstruncate.py
```diff
@@ -26,7 +26,7 @@ class BodyTooLongException(Exception):
     pass
 
 
-def is_too_long(payload, max_length=2048):
+def is_too_long(payload: Dict[Any, Any], max_length: int = 2048) -> bool:
```
babolivier, line 29: Looks like the type annotation you want is `dict` here.

babolivier (changes requested):

sygnal/apnstruncate.py
babolivier: Now that you've changed that with https://github.com/matrix-org/sygnal/pull/225/commits/8ae8f5212c77ab3899cf5577e8ebfb5f45b8493c you'll also want to remove the import to `typing.Dict` on line 18 since it's not used anymore.

kibablu: Thank you, I have just removed it after seeing my test fails.

babolivier (approved):

babolivier:
Thanks for your contribution! :)

sygnal/apnstruncate.py
kibablu: @babolivier is there a difference between `typing.Dict` and `dict` on python versions? 
while running `mypy --strict sygnal/ tests/ setup.py` I'm getting an error :point_down: 
dict
I'm running `python 3.8`
https://github.com/matrix-org/sygnal/blob/656c0ead8750a5630080a6bd0c63b1bae52a478f/sygnal/apnstruncate.py#L29

PR #227, opened 2021-06-01
Improve static type checking

kibablu:
Add type annotation to truncate Function on apnstruncate.py file
as part of the #201 issue

### DIFF
--- a/sygnal/apnstruncate.py
+++ b/sygnal/apnstruncate.py
@@ -15,7 +15,7 @@
 # Copied and adapted from
 # https://raw.githubusercontent.com/matrix-org/pushbaby/master/pushbaby/truncate.py
 import json
-from typing import Any, List, Tuple, Union
+from typing import Any, Dict, List, Tuple, Union
 
 
 def json_encode(payload: Any) -> bytes:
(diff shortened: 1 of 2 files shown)


sygnal/apnstruncate.py
```diff
@@ -38,7 +38,7 @@ def is_too_long(payload: dict, max_length: int = 2048) -> bool:
     return len(json_encode(payload)) > max_length
 
 
-def truncate(payload, max_length=2048):
+def truncate(payload: Dict[Any, Any], max_length: int = 2048) -> Dict[str, Any]:
```
callahad, line 41:
```suggestion
def truncate(payload: Dict[str, Any], max_length: int = 2048) -> Dict[str, Any]:
```

If we're returning a `Dict[str, Any]` we should probably restrict the inputs to be the same

callahad (changes requested):
Looks good aside from one small tweak requested

sygnal/apnstruncate.py
kibablu: Thank you :smiley:  I have made the changes

callahad:
🎉  Thank you so much!

callahad (approved):

PR #234, opened 2021-07-13
Explore removing DB from Sygnal

H-Shay:
As a first step in the exploration of possibly dropping the DB from Sygnal, in this PR I put all the DB functionality behind a config variable check. This means that in one of the two functions that actually interact with the DB, I added a check at the beginning which calls another function with the DB calls dropped if the DB is not being used (per the config variable).

The other function only called the DB once so I just put that call behind an `if` statement in the function. Note that this formulation still instantiates a DB, we just don't ever call it if the config variable is `false`. 

I spent some time trying to remove the DB entirely if the config variable was `false`, but that involved creating and forcing a choice between a regular `gmc_pushkin` and a whole new class of `gmc_pushkin` constructed without a DB and that began to seem excessive given that hopefully this is just a transitory step towards dropping the DB entirely. However, if that would be the right way to do it I am happy to give it a shot. 

I also had to modify the `gmc_tests` a little to give them a pre-loaded config variable, and to skip two tests that were specific to the DB function if the config variable is `false`. The tests, sygnal itself, and the sygnal.yaml.sample are all set to default to the config variable being `false` (although I did test them with it set to `true` also to ensure it worked on both settings). 

I did also spin up an instance of Sygnal to ensure that the new config variable was being picked up out of the `sygnal.yaml` file, but I was unable to figure out how to write a test to do that, as all the test class instances of Sygnal did not pull from the `sygnal.yaml` file and I got a little bit in the weeds trying to figure out how to do that.


reivilibre:
Is this now obsolete since #236 ?

reivilibre:
we've decided this is superseded by #236

PR #242, opened 2021-08-05
Add testing for old dependencies

reivilibre:
We should be honest in our list of requirements. This PR adds a test pass for old deps, and will also bump up deps slightly to match the reality of what we've secretly been depending on all along.

Reasons for bumps made.

```
The user requested cryptography==2.1.4
    service-identity 18.1.0 depends on cryptography
    pywebpush 1.13.0 depends on cryptography>=2.6.1
```


dependencies.py
```diff
@@ -0,0 +1,28 @@
+INSTALL_REQUIRES = [
```
richvdh, line 1: could this be inside setup.py? and have old-deps read that rather than this? You might need an `if __name__ == '__main__'` in setup.py.

If we're having a separate file, we need license headers and comments about what it's for, but I think it's easier to roll it into setup.py.

scripts-dev/lint.sh
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env bash
```
richvdh, line 1: this should probably have been a separate PR...

scripts-dev/lint.sh
```diff
@@ -0,0 +1,101 @@
...
+  # If we were not asked to lint changed files, and no paths were found as a result,
+  # then lint everything!
+  if [[ -z ${files+x} ]]; then
+    # Lint all source code files and directories
+    # Note: this list aims to mirror the one in tox.ini
+      files=(
+          "sygnal" "tests"
+          "scripts-dev/old_deps.py"
+          "setup.py" "dependencies.py"
+      )
```
richvdh, line 88: the indentation is bogus here.

richvdh (changes requested):

scripts-dev/lint.sh
reivilibre: you're not wrong, but so is Synapse's :stuck_out_tongue_winking_eye:

dependencies.py
reivilibre: apparently yes it can be :) with if..main.

I had trouble with it earlier and saw Synapse splits it out, so thought there must have been a good reason for it...

scripts-dev/lint.sh
reivilibre: you're right... (is this your script?) I love it too much, I couldn't resist (I also notice it's grown the difference option since last time I looked at it :))

will move over

scripts-dev/old_deps.py
```diff
@@ -0,0 +1,32 @@
+import os
```
callahad, line 1: Missing shebang?

.github/workflows/pipeline.yml
```diff
@@ -35,3 +36,18 @@ jobs:
...
+    name: Unit tests (old dependencies)
+    needs: [ check-code-style, check-types-mypy ]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.7"
+      - name: Install old dependencies
+        run: python scripts-dev/old_deps.py | pip install -r /proc/self/fd/0
```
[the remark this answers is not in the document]
callahad: ...should `scripts-dev/old_deps.py` be executable, and then we could just `pip install -r <(./scripts-dev/old_deps.py)` ?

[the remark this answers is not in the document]
reivilibre: ah yes, forgot bash can do that

setup.py
```diff
@@ -17,53 +17,57 @@
...
-            "flake8==3.9.0",
-            "isort~=5.0",
-            "mypy==0.812",
-            "mypy-zope==0.3.0",
-            "tox",
-        ]
-    },
-    long_description=read("README.rst"),
-)
+def read_file(path_segments):
```
richvdh, line 53: it's a bit of a shame to lose the type annotations from the old `read`.

scripts-dev/old_deps.py
```diff
@@ -0,0 +1,33 @@
...
+This script outputs a list of requirements, which, if fed to pip install -r,
+will install the oldest versions of the dependencies specified by the
+setup.py file in the repository root.
+"""
+
+
+here = os.path.abspath(os.path.dirname(__file__))
+
+
+def read_file(path_segments):
```
richvdh, line 14: type annotations please.

scripts-dev/old_deps.py
```diff
@@ -0,0 +1,33 @@
...
+    """Read a file from the package. Takes a list of strings to join to
+    make the path"""
```
richvdh, line 16:
```suggestion
    """Read a file from the package. 
    
    Params:
        path_segments: a list of strings to join to make the path
    """
```

scripts-dev/old_deps.py
```diff
@@ -0,0 +1,33 @@
...
+    file_path = os.path.join(here, *path_segments)
+    with open(file_path) as f:
+        return f.read()
+
+
+def exec_file(path_segments):
```
richvdh, line 22: type annotations please

scripts-dev/old_deps.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
```
richvdh, line 1: we ought to have a license header here, really.

tox.ini
```diff
@@ -23,9 +23,9 @@ commands =
 [testenv:check_codestyle]
 
 commands =
-  flake8 sygnal/ tests/ setup.py
-  black --check --diff sygnal/ tests/ setup.py
-  isort --check-only --diff sygnal/ tests/ setup.py
+  flake8 sygnal/ tests/ setup.py scripts-dev/old_deps.py
+  black --check --diff sygnal/ tests/ setup.py scripts-dev/old_deps.py
+  isort --check-only --diff sygnal/ tests/ setup.py scripts-dev/old_deps.py
```
richvdh, line 28: just do the whole of scripts-dev?

richvdh (changes requested):

scripts-dev/old_deps.py
```diff
@@ -1,5 +1,21 @@
 #!/usr/bin/env python
+
```
richvdh, line 2:
```suggestion
#
```

... just for consistency with everywhere else.

richvdh (approved):
lgtm

PR #249, opened 2021-08-10
Build and push docker images using GitHub Actions

### LINKED ISSUE #245: Docker build should run in CI
CI should generate docker images and push them (for releases) to Docker Hub

reivilibre:
Closes #245.

### DIFF
--- a/.github/workflows/docker_push.yml
+++ b/.github/workflows/docker_push.yml
@@ -34,21 +34,14 @@ jobs:
           username: ${{ secrets.DOCKER_HUB_USERNAME }}
           password: ${{ secrets.DOCKER_HUB_TOKEN }}
 
-      - name: Calculate docker image tag
+      - name: Calculate docker image tags
         id: set-tag
-        run: |
-          case "${GITHUB_REF}" in
-              refs/heads/master|refs/heads/main)
-                  tag=latest
-                  ;;
-              refs/tags/*)
-                  tag=${GITHUB_REF#refs/tags/}
-                  ;;
-              *)
-                  tag=${GITHUB_SHA}
-                  ;;
-          esac
-          echo "::set-output name=tag::$tag"
+        uses: docker/metadata-action@v3
+        with:
+          images: reivilibre/sygnal
+          tags: |
+            type=ref,event=tag
+            type=sha,enable=true,priority=100,prefix=,suffix=,format=long
 
       # we explicitly check out the repository (and use `context: .` in buildx)
       # because we need to preserve the git metadata so that setuptools_scm
@@ -63,6 +56,6 @@ jobs:
           context: .
           push: true
           labels: "gitsha1=${{ github.sha }}"
-          tags: "reivilibre/sygnal:${{ steps.set-tag.outputs.tag }}"
+          tags: "${{ steps.set-tag.outputs.tags }}"
           file: "docker/Dockerfile"
           platforms: linux/amd64,linux/arm64
--- a/.github/workflows/docker_push.yml
+++ b/.github/workflows/docker_push.yml
@@ -42,7 +42,7 @@ jobs:
         id: set-tag
         uses: docker/metadata-action@v3
         with:
-          images: reivilibre/sygnal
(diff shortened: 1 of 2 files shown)


reivilibre:
https://hub.docker.com/repository/registry-1.docker.io/reivilibre/sygnal/tags?page=1&ordering=last_updated
https://github.com/reivilibre/sygnal

This seems to work

.github/workflows/docker_push.yml
```diff
@@ -0,0 +1,64 @@
+# GitHub actions workflow which builds and publishes the docker images.
+
+name: Build and push docker images
+
+on:
+  push:
+    tags: ["v*"]
+    # uncomment this to enable all pushes to main
+    #branches: [ master, main ]
```
anoadragon453, line 9: I think `latest` on our docker images always just points to the latest release. Not sure if we'd ever push on the `main` branch.

Though you mention re-enabling building docker images on main below - did we used to do that?

changelog.d/249.docker
```diff
@@ -0,0 +1 @@
+Automatically build the Docker image and push it to Docker Hub using GitHub Actions.
```
anoadragon453, line 1:
```suggestion
Automatically build the Docker image for each release and push it to Docker Hub using GitHub Actions.
```

.github/workflows/docker_push.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Log in to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_TOKEN }}
+
+      # The following line can be used to tag builds with the SHA hash of the commit.
+      # This would be useful if we wanted to re-enable building for all commits on `main`.
+      # type=sha,enable=true,priority=100,prefix=,suffix=,format=long
```
anoadragon453, line 40: This is what would generate the following tag I assume?

image

anoadragon453 (commented):
Looking great! Just a couple questions below.

.github/workflows/docker_push.yml
reivilibre: I ripped off https://github.com/matrix-org/synapse/blob/develop/.github/workflows/docker.yml which seems like if you push to master on Synapse, you will get a new build.

I'm not really sure why this is, so I took it out for Sygnal.

.github/workflows/docker_push.yml
reivilibre: Yeah, that's right. Reckon I should just cut it out entirely?

.github/workflows/docker_push.yml
anoadragon453: Given we don't have anything other than releases on [our dockerhub](https://hub.docker.com/r/matrixdotorg/synapse/tags?page=1&ordering=last_updated), I believe that that workflow both tests docker builds as well as publishes them (if the branch has an associated release tag).

This isn't necessary in Sygnal, as we already have a workflow for testing docker builds.

.github/workflows/docker_push.yml
anoadragon453: yeah, I'd cut it.

anoadragon453 (dismissed):
lgtm once the above is sorted.

bradtgmurray:
Now that this is resolved, can we pretty please get a point release tagged and pushed? Looking forward to trying out the db-less sygnal release, but unfortunately the container isn't available on dockerhub. https://hub.docker.com/r/matrixdotorg/sygnal/tags

reivilibre:
> Now that this is resolved, can we pretty please get a point release tagged and pushed? Looking forward to trying out the db-less sygnal release, but unfortunately the container isn't available on dockerhub. https://hub.docker.com/r/matrixdotorg/sygnal/tags

That's the plan, yup, sorry for taking my sweet time with it :)

reivilibre:
@bradtgmurray 0.10.1 released -- go and have fun! :D

Issue #230, opened 2021-06-10
App type "gcm" in the config is confusing as we exclusively use FCM


anoadragon453:
We switched a long while ago from Google Cloud Messaging (GCM) to Firebase Cloud Messaging (FCM). GCM [essentially became FCM](https://developers.google.com/cloud-messaging/android/android-migrate-fcm).

Thus APNs, **FCM** and WebPush are the only push backends we currently support in Sygnal. However, it can be confusing that people running the bridge need to specify `gcm` instead of `fcm`, as it's not clear which one is actually being used.

It may be nice to allow both `gcm` and `fcm` as a configurable app type, and to document that both mean "use fcm".

At the same time we can deprecate the use of `gcm` as a type in the config, and eventually drop it.

PR #251, opened 2021-08-20
Clarify use of 'FCM' in app type and refactor to use 'FCM'

H-Shay:
Addresses issue #230. This PR allows for both `gcm` and `fcm` to be used as app types, warns that the app type `gcm` will soon be depreciated, and refactors sygnal to clarify that we are using Firebase Cloud Messaging exclusively.


H-Shay:
Totally baffled as to why the changelog check is failing.

README.rst
```diff
@@ -63,11 +63,14 @@ apns
...
-gcm
-  This sends messages via Google/Firebase Cloud Messaging (GCM/FCM) and hence can be used
+fcm
+  This sends messages via Firebase Cloud Messaging (FCM) and hence can be used
   to deliver notifications to Android apps. It expects the 'api_key' parameter
   to contain the 'Server key', which can be acquired from Firebase Console at:
   ``https://console.firebase.google.com/project/<PROJECT NAME>/settings/cloudmessaging/``
+  Please note that the legacy key 'gcm' can be used here in place of fcm, however,
+  both keys refer to Firebase Cloud Messaging and the gcm key will soon be
+  depreciated and dropped in future releases.
```
anoadragon453, line 73:
```suggestion
  deprecated and dropped in future releases.
```

anoadragon453, line 73: We probably want to deprecate the key in this release, and then remove it some time in the future, rather than warning about deprecation.

changelog.d/251.misc
```diff
@@ -0,0 +1 @@
+Addresses #230, refactors sygnal to use fcm as app type rather than gcm, and warns in documents about the future depreciation of gcm as an app type.
```
anoadragon453, line 1:
```suggestion
Addresses #230, refactors sygnal to use fcm as app type rather than gcm, and warns in documents about the future deprecation of gcm as an app type.
```

sygnal.yaml.sample
```diff
@@ -198,13 +198,16 @@ apps:
   #  # The default is 'production'. Uncomment to use the sandbox instance.
   #  #platform: sandbox
 
-  # This is an example GCM/FCM push configuration.
+  # This is an example FCM/GCM push configuration. Note that the gcm key is a legacy
```
anoadragon453, line 201:
```suggestion
  # This is an example FCM/GCM push configuration. Note that the gcm type is a legacy
```

sygnal.yaml.sample
```diff
@@ -198,13 +198,16 @@ apps:
...
+  # setting as we now use firecloud messaging (fcm) exclusively. Setting gmc as the
```
anoadragon453, line 202:
```suggestion
  # setting as we now use firecloud messaging (fcm) exclusively. Setting gcm as the
```

sygnal.yaml.sample
```diff
@@ -198,13 +198,16 @@ apps:
...
+  # app type is still accepted but understand that under the hood we are using fcm.
+  # The gcm app type will be depreciated in future releases.
```
anoadragon453, line 204:
```suggestion
  # The gcm app type will be deprecated in future releases.
```

sygnal/fcmpushkin.py
```diff
@@ -37,38 +37,38 @@
...
-    "sygnal_active_gcm_requests", "Number of GCM requests in flight"
+    "sygnal_active_fcm_requests", "Number of FCM requests in flight"
 )
 
 RESPONSE_STATUS_CODES_COUNTER = Counter(
-    "sygnal_gcm_status_codes",
-    "Number of HTTP response status codes received from GCM",
+    "sygnal_fcm_status_codes",
+    "Number of HTTP response status codes received from FCM",
     labelnames=["pushkin", "code"],
```
anoadragon453, line 58: Note that modifying these will likely break existing metrics dashboards. Thus this (as well as the span tag name changes below) will need to be detailed in the changelog as well so sysadmins aren't caught off guard :slightly_smiling_face:

anoadragon453 (changes requested):
> Totally baffled as to why the changelog check is failing.

Looks like the CI is finding the changelog at `<checkout dir>/sygnal/changelog.d/251.misc`, rather than `<checkout dir>/changelog.d/251.misc`, which its expecting.

The changelog CI code was [just recently updated](https://github.com/matrix-org/sygnal/pull/251), and the changelog location in your PR looks correct. I might suggest trying to merge in the `main` branch and see if that works.

sygnal/fcmpushkin.py
H-Shay: So afaict my main branch is up to date with the latest sygnal changes, and this fork as well-attempting to merge from main just gives an "Already up to date" message, yet the changelog check is still failing. I can't tell if I am missing something simple here!

reivilibre:
the changelog issue is odd. It was shown to work in the PR that added it :|. I can't see what's wrong with it or why it would act differently in either case.

reivilibre:
it seems like there's an issue with the newsfile checker *when used with a fork rather than a branch*, as I found out by experimenting with #253 (and #252).
I'll look into this, sorry for the hassle

sygnal/fcmpushkin.py
anoadragon453: Looks like it's passing now! (Thanks @reivilibre!)

changelog.d/251.misc
```diff
@@ -0,0 +1,15 @@
+Addresses #230, refactors sygnal to use fcm as app type rather than gcm, and warns in documents about the future deprcation of gcm as an app type. Due to these changes, some prometheus metrics names were updated to refer to Firebase Cloud Messaging (fcm) instead of the legacy Google Cloud Messaging (GCM) services. You can find a list of the changed metric names below:
```
anoadragon453, line 1:
```suggestion
Addresses #230, refactors sygnal to use fcm as app type rather than gcm, and warns in documents about the future deprecation of gcm as an app type. Due to these changes, some prometheus metrics names were updated to refer to Firebase Cloud Messaging (fcm) instead of the legacy Google Cloud Messaging (GCM) services. You can find a list of the changed metric names below:
```

Additionally, there's still a few places which say "will deprecate" rather than "is deprecated" which should be updated :slightly_smiling_face:

changelog.d/251.misc
```diff
@@ -0,0 +1,15 @@
...
+    sygnal_active_gcm_requests -> sygnal_active_fcm_requests
+    sygnal_gcm_status_codes -> sygnal_fcm_status_codes
+
+Additionally, tag identifiers in OpenTracing Spans were also updated:
+
+    gcm_error -> fcm_error
+    gcm_num_devices -> fcm_num_devices
+    gcm_num_failed -> fcm_num_failed
+
+If you make use of either Prometheus (and thus Grafana, etc) metrics or use OpenTracing with Sygnal, please ensure your queries are up to date.
```
anoadragon453, line 15: Thanks for including this information! Note that newsfragments (this file) should just be one line that gives a brief summary. For the call out of the deprecation, that should just be placed at the top of the `CHANGELOG.md` file as part of this PR (rather than the newsfragment).

anoadragon453 (changes requested):
Looks good to me other than the minor bits below!

sygnal.yaml.sample
```diff
@@ -198,13 +198,16 @@ apps:
   #  # The default is 'production'. Uncomment to use the sandbox instance.
   #  #platform: sandbox
 
-  # This is an example GCM/FCM push configuration.
+  # This is an example FCM/GCM push configuration. Note that the gcm type is a legacy
+  # setting as we now use Firebase Cloud Messaging (fcm) exclusively. Setting gcm as the
+  # app type is still accepted but understand that under the hood we are using fcm.
+  # The gcm app type is deprecated.
```
reivilibre, line 204:
```suggestion
  # This is an example FCM push configuration.
  # FCM was formerly known as GCM.
  # The deprecated app type `gcm` is equivalent but will be removed from a future Sygnal release.
```

sygnal/fcmpushkin.py
```diff
@@ -37,38 +37,38 @@
...
 RETRY_DELAY_BASE = 10
 MAX_BYTES_PER_FIELD = 1024
 
 # The error codes that mean a registration ID will never
 # succeed and we should reject it upstream.
 # We include NotRegistered here too for good measure, even
-# though gcm-client 'helpfully' extracts these into a separate
+# though fcm-client 'helpfully' extracts these into a separate
 # list.
 BAD_PUSHKEY_FAILURE_CODES = [
```
reivilibre, line 73:
```suggestion
BAD_PUSHKEY_FAILURE_CODES = [
```

`gcm-client` was a library for using GCM, so renaming the reference to fcm-client doesn't make sense anymore.

However, gcm-client isn't even in use, so this comment is probably _very_ crufty — hence I think it should just be removed altogether.

reivilibre (commented):
some little tweaks I've spotted. I think anoa will be in charge of giving this the tick though still?

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-This release deprecates the use of gcm as an app type and refactors Sygnal to use fcm as an app type exclusively. Due to these changes, some prometheus metrics names were updated to refer to Firebase Cloud Messaging (fcm) instead of the legacy Google Cloud Messaging (GCM) services. You can find a list of the changed metric names below:
+This release deprecates the use of 'gcm' as an app type and refactors Sygnal to use 'fcm' as an app type exclusively. Due to these changes, some Prometheus metrics' names were updated to refer to Firebase Cloud Messaging (FCM) instead of the legacy Google Cloud Messaging (GCM) services. You can find a list of the changed metric names below:
```
reivilibre, line 1:
```suggestion
This release deprecates the use of `gcm` as an app type and refactors Sygnal to use `fcm` as an app type exclusively. Due to these changes, some Prometheus metrics' names were updated to refer to Firebase Cloud Messaging (FCM) instead of the legacy Google Cloud Messaging (GCM) services. You can find a list of the changed metric names below:
```

(I would probably have said to use \` as it will show it in code blocks, since it's a literal from the config file)

changelog.d/251.misc
```diff
@@ -1 +1 @@
-Addresses #230, refactors sygnal to use fcm as app type rather than gcm, and warns in documents about the deprecation of gcm as an app type.
\ No newline at end of file
+Rename the 'gcm' app type to 'fcm' to reduce confusion (Google Cloud Messaging changed name to Firebase Cloud Messaging in 2016). The app type 'gcm' is deprecated and will be removed in a future release.
```
reivilibre, line 1: (also here? It winds up in the markdown document.)

```suggestion
Rename the `gcm` app type to `fcm` to reduce confusion (Google Cloud Messaging changed name to Firebase Cloud Messaging in 2016). The app type `gcm` is deprecated and will be removed in a future release. 
```

reivilibre (commented):
maybe anoa will have opinions about ' vs `.

callahad:
Without looking at this too closely, do we have opinions about referring to this as "FCM" versus something more specific to the FCM API we implement?

Namely, Sygnal supports the "legacy HTTP" API, not the newer "HTTP v1" API (cite: [Migrate from legacy HTTP to HTTP v1](https://firebase.google.com/docs/cloud-messaging/migrate-v1))

Concretely this would probably mean using `fcm_legacy` or similar instead of just `fcm` so that we have room for `fcm_v1` later.

reivilibre:
> Without looking at this too closely, do we have opinions about referring to this as "FCM" versus something more specific to the FCM API we implement?
> 
> Namely, Sygnal supports the "legacy HTTP" API, not the newer "HTTP v1" API (cite: [Migrate from legacy HTTP to HTTP v1](https://firebase.google.com/docs/cloud-messaging/migrate-v1))
> 
> Concretely this would probably mean using `fcm_legacy` or similar instead of just `fcm` so that we have room for `fcm_v1` later.

On one hand, I don't see much advantage, unless we intend to support both separately? They may be two different APIs, but they are both APIs onto the same service. We never had an app type change when we moved over from the old, binary-protocol APNS to the HTTP/2 APNS, for example. Renaming it to something with 'legacy' in the name is kind of unfortunate when it's the _'gcm'_ option that's the legacy option.

On the other hand, perhaps we will migrate to the v1 API one day (if there's a use case for it). I can't tell if admins will need to make any config changes then or not (it seems like the format of access tokens might have changed, so perhaps?). If that's the case, I can feel _someone_ wanting to say that we should honour both options (:P) for some time whilst admins migrate (although I suppose nobody is forcing them to upgrade Sygnal until they're ready to amend the configuration).

anoadragon453:
Having `fcm_legacy` without a `fcm` sounds a bit confusing. I'd suggest we either just do `fcm_v1` when the time comes, or simply add an additional option under the pushkin config to choose the API, while still keeping an app type of `fcm`.

CHANGELOG.md
anoadragon453: From my experience, `gcm` is what I would use if `gcm` were the name of a config option key. However, I would write "gcm" if I were referring to a config option *value*.

Both are fine though, I don't think it needs to be debated too much.

anoadragon453 (commented):
I have no more suggestions other than to resolve the open discussions above :slightly_smiling_face:

rltas:
What about supporting both type: gcm and type: fcm and introducing a new parameter `server_key` for the latter?
Then for a transition period, existing configs keep working with the legacy API without config changes:
type: gcm
api_key: your_api_key_for_gcm
But if you're going with fcm, it would have to be:
type: fcm
server_key: your_server_key_for_gcm
This would be more proper (less confusing) naming anyway.
Then when #313 is ready, issue deprecation warnings on startup for a transition period when server_key is used, before finally removing the old code altogether, because Google will remove the legacy API at some point.
This would be a good process for those who have to maintain Sygnal (admins), because they're usually not the ones doing the Google/app stuff.
Because this change has been delayed for quite some time, while Matrix gained quite a bit of popularity, many new setups unfortunately now use type: gcm and have to be given some head notice.
When using the ansible playbook to setup a Matrix stack, this can be elegantly mitigated anyway.

Issue #255, opened 2021-09-13
Investigate IndexError in gcm's `_dispatch_notification_unlimited`


DMRobertson:
(internal users only:) https://sentry.matrix.org/sentry/sygnal-matrixorg/issues/201296/?query=is%3Aunresolved

The list comprehension that builds `pushkeys` might produce an empty list, in which case we get the index error as reported on sentry.

This logic seems to have been here since eeea1a71a1aa70e629a05f5d85e6f7d7c3406c7a in Fed 2015. The first sentry report is from Mar 6, 2021.

Questions:

- is it sensible for this list comprehension to return an empty list?
- what should we do if so
- did something change in March to trigger us seeing this?

reivilibre:
I saw this pop up in my e-mails.

It doesn't make sense for someone to send a push notification to zero pushkeys but I guess someone has accidentally done that anyway.

It would probably be sensible to catch this case and either 200 it without doing anything (OK! I've done nothing as you requested) or 400 Bad Request it.

Technically the spec doesn't say that there should be at least 1 device, but the call doesn't make any sense if you don't specify a device, so either option seems reasonable to me (though perhaps 400 Bad Request is a clearer error for when an implementation gets it wrong and is trying to figure out why push doesn't work).

DMRobertson:
Assuming that the pushkey is part of the HTTP request, I agree that 400 "You messed up" is the right response here.

But having taken another look I'm not sure how we actually end up in this situation. My reading: for each device `d in notif.devices` we check that exactly one pushkin exists matching its the devices's `app_id`.

https://github.com/matrix-org/sygnal/blob/ff1e98ed494cf388606c4dd33eb38a0f67593cc2/sygnal/http.py#L243-L269

We then dispatch to the _single device `d`_ via `dispatch_notification` and `_dispatch_notification_unlimited`. By the time we're in the latter function, I think it should always be the case that the `device` argument to `_dispatch_notification_unlimited` should always satisfy  `device.app_id == self.name`. And so I think this list comprehension should never be empty?

https://github.com/matrix-org/sygnal/blob/d52568e248a00015e6759d354bc80205f3069709/sygnal/gcmpushkin.py#L300-L305

If we are dispatching to just a single device, I don't understand why we are even looking at `n.devices` in the first place (and I couldn't see other pushkins doing something similar). Can't we just use the single pushkey `device.pushkey`, made into a list of length 1 if needbe?

DMRobertson:
[...]

PR #256, opened 2021-09-16
Add .git-blame-ignore-revs

DMRobertson:
Configurable with
```shell
git config --local blame.ignoreRevFile=.git-blame-ignore-revs
```

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Update to use Twisted and python 3.7 (#42)
+b784e46f3a22700efac8560d0ab0230e7c3f955f


reivilibre:
Can you explain what this does?

DMRobertson:
Sure.

Git 2.23 apparently added the option to "ignore" certain commits when working out which commit to `git blame` for a particular line. See [this stackoverflow answer](https://stackoverflow.com/a/57129540) ; there's also a mention in the [Git 2.23 changelog](https://public-inbox.org/git/xmqqy2zszuz7.fsf@gitster-ct.c.googlers.com/).

The goal is to filter out commits from `blame` that are fluffy, non-semantic changes. Think mass reformatting, or some kind of mass renaming that shouldn't have an impact on runtime behaviour.

We use this [in synapse](https://github.com/matrix-org/synapse/blob/develop/.git-blame-ignore-revs) too.

Adding this file by itself does nothing; the behaviour is opt-in. One either passes `--ignore-revs-file=FILE` to a `git blame` invocation, or otherwise uses `git config` to opt into this behaviour automatically.

DMRobertson:
Hmmm. Taking a look at [the commit that I wanted to ignore](https://github.com/matrix-org/sygnal/commit/b784e46f3a22700efac8560d0ab0230e7c3f955f) and [the corresponding PR](https://github.com/matrix-org/sygnal/pull/42), it looks like a biggggg squash with both  formatting and functional changes. I wanted to filter out 

> * Reformat code using `black`

but there are lots of other things in there that we probably don't want to ignore, like

> * Add Sentry integration
> * Add Prometheus support 
>  Deal with errors by reporting 500 or 502 as appropriate when things g… 
> Generate fresh UUIDs for APNs notifications as they are not just opaque 

Given that, maybe we should drop this change

reivilibre:
Thanks, interesting to know that this tool exists, though yeah, from the description I'm not sure this commit (the Sygnal 'rewrite') is the best use for it, since does that mean you won't be able to see a git blame for lines that go back to Sygnal-rewrite's 'birth'?

There *are* a few lines that still bleed through (mostly where the logic was sound and not dependent on either gevent or Twisted), but largely I don't know if it's so bad that most things wind up pointing to the birth of the Sygnal rewrite.

I'm wondering what specifically motivated this?

clokep:
I took a quick look and I don't think we want to drop this commit since it has functional changes too.

DMRobertson:
> I'm wondering what specifically motivated this?

I was browsing around in `git blame` to see how often the GCM pushkin had been changed. I wanted the soft-context of "what behaviour might have changed to cause #255?". I kept seeing this particular commit and saw 

> * Reformat code using `black`

in its description, so assumed it was uninteresting for this kind of commit archaeology.

Having taken another look, that assumption now seems wrong (and Patrick concurs). So let's close this off for now.

PR #264, opened 2021-11-04
Type hints for apnstruncate, apnspushkin, notifications, and gcmpushkin

H-Shay:
Here is my first try at adding some type hints to the below files in Sygnal, and making very minor modifications to the code to silence some mypy errors (like `assert is not None`). 
A very small handful of errors remain (less than 4 I think?) across the files I changed but silencing them seemed like it might require more modification to the code, and I wanted to make sure that everything I had here was correct before I started tackling those. 
I think that these are mostly correct but I find typing to be a little difficult so it is entirely possible I have gotten some things wrong.


sygnal/apnspushkin.py
```diff
@@ -237,7 +254,9 @@ async def _dispatch_request(self, log, span, device, shaved_payload, prio):
                         f"{response.status} {response.description}"
                     )
 
-    async def _dispatch_notification_unlimited(self, n, device, context):
+    async def _dispatch_notification_unlimited(
+        self, n: Notification, device: Device, context: NotificationContext
+    ) -> Optional[list]:
```
reivilibre, line 259:
```suggestion
    ) -> List[str]:
```
in general, avoid writing bare `list` as a type, and also try to specify the parameters if you can :)

sygnal/apnspushkin.py
```diff
@@ -180,7 +189,14 @@ def _report_certificate_expiration(self, certfile):
...
 
-    async def _dispatch_request(self, log, span, device, shaved_payload, prio):
+    async def _dispatch_request(
+        self,
+        log: NotificationLoggerAdapter,
+        span: Span,
+        device: Device,
+        shaved_payload: Dict[str, Any],
+        prio: int,
+    ) -> Optional[List]:
```
reivilibre, line 199:
```suggestion
    ) -> Optional[List[str]]:
```

sygnal/apnspushkin.py
```diff
@@ -250,14 +269,16 @@ async def _dispatch_notification_unlimited(self, n, device, context):
...
+                    n, device
+                )
             else:
                 payload = self._get_payload_full(n, device, log)
 
             if payload is None:
                 # Nothing to do
                 span_parent.log_kv({logs.EVENT: "apns_no_payload"})
-                return
+                return None
```
reivilibre, line 281:
```suggestion
                return []
```
I'm in favour of avoiding confusion in the return type; nothing wrong with returning None but the empty list makes sense here I think

sygnal/apnspushkin.py
```diff
@@ -273,6 +294,7 @@ async def _dispatch_notification_unlimited(self, n, device, context):
                     with self.sygnal.tracer.start_span(
                         "apns_dispatch_try", tags=span_tags, child_of=span_parent
                     ) as span:
+                        assert shaved_payload is not None
```
reivilibre, line 297: This feels like something that we should avoid being necessary

reivilibre: Actually yes: why *is* this necessary :o

sygnal/apnstruncate.py
```diff
@@ -38,7 +38,9 @@ def is_too_long(payload: Dict[Any, Any], max_length: int = 2048) -> bool:
     return len(json_encode(payload)) > max_length
 
 
-def truncate(payload: Dict[str, Any], max_length: int = 2048) -> Dict[str, Any]:
+def truncate(
+    payload: Dict[str, Any], max_length: int = 2048
+) -> Optional[Dict[str, Any]]:
```
reivilibre, line 43: Why did this become optional? I can't see at first glance

sygnal/apnstruncate.py
```diff
@@ -85,7 +87,7 @@ def truncate(payload: Dict[str, Any], max_length: int = 2048) -> Dict[str, Any]:
     return payload
 
 
-def _choppables_for_aps(aps):
+def _choppables_for_aps(aps: dict) -> List[Union[Tuple[str], Tuple[str, int]]]:
```
reivilibre, line 90: I would probably narrow this a little
```suggestion
def _choppables_for_aps(aps: Dict[str, Any]) -> List[Union[Tuple[str], Tuple[str, int]]]:
```

sygnal/apnstruncate.py
```diff
@@ -102,7 +104,7 @@ def _choppables_for_aps(aps):
     return ret
 
 
-def _choppable_get(aps, choppable):
+def _choppable_get(aps: dict, choppable: Union[Tuple[str], Tuple[str, int]]):
```
reivilibre, line 107: also here
```suggestion
def _choppable_get(aps: Dict[str, Any], choppable: Union[Tuple[str], Tuple[str, int]]):
```

sygnal/apnstruncate.py
```diff
@@ -111,7 +113,9 @@ def _choppable_get(aps, choppable):
         return aps["alert"]["loc-args"][choppable[1]]
 
 
-def _choppable_put(aps, choppable, val):
+def _choppable_put(
+    aps: dict, choppable: Union[Tuple[str], Tuple[str, int]], val: str
```
reivilibre, line 117:
```suggestion
    aps: Dict[str, Any], choppable: Union[Tuple[str], Tuple[str, int]], val: str
```

sygnal/apnstruncate.py
```diff
@@ -120,7 +124,7 @@ def _choppable_put(aps, choppable, val):
         aps["alert"]["loc-args"][choppable[1]] = val
 
 
-def _longest_choppable(aps):
+def _longest_choppable(aps: dict) -> Union[Tuple[str], Tuple[str, int], None]:
```
reivilibre, line 127:
```suggestion
def _longest_choppable(aps: Dict[str, Any]) -> Union[Tuple[str], Tuple[str, int], None]:
```

sygnal/gcmpushkin.py
```diff
@@ -109,8 +119,8 @@ def __init__(self, name, sygnal, config):
             )
 
         self.http_pool = HTTPConnectionPool(reactor=sygnal.reactor)
-        self.max_connections = self.get_config(
-            "max_connections", DEFAULT_MAX_CONNECTIONS
+        self.max_connections = cast(
+            int, self.get_config("max_connections", DEFAULT_MAX_CONNECTIONS)
         )
```
reivilibre, line 124: This will conceal problems if someone doesn't specify the right thing, better to check it:
```suggestion
        max_connections_raw = self.get_config("max_connections", DEFAULT_MAX_CONNECTION)
        if not isinstance(max_connections_raw, int):
            raise PushkinSetupError("max_connections (if specified) must be an int")
        self.max_connections = max_connections_raw
```

sygnal/gcmpushkin.py
```diff
@@ -134,14 +144,16 @@ def __init__(self, name, sygnal, config):
         # Use the fcm_options config dictionary as a foundation for the body;
         # this lets the Sygnal admin choose custom FCM options
         # (e.g. content_available).
-        self.base_request_body: dict = self.get_config("fcm_options", {})
+        self.base_request_body: dict = cast(dict, self.get_config("fcm_options", {}))
         if not isinstance(self.base_request_body, dict):
             raise PushkinSetupException(
                 "Config field fcm_options, if set, must be a dictionary of options"
             )
```
reivilibre, line 151: again, this is awkward (and it's already being checked) but using `cast` should be avoided
```suggestion
        base_request_body_raw = self.get_config("fcm_options", {})
        if not isinstance(base_request_body_raw, dict):
            raise PushkinSetupException(
                "Config field fcm_options, if set, must be a dictionary of options"
            )
        self.base_request_body = base_request_body_raw
```

sygnal/gcmpushkin.py
```diff
@@ -190,7 +204,15 @@ async def _perform_http_request(self, body, headers):
...
-    async def _request_dispatch(self, n, log, body, headers, pushkeys, span):
+    async def _request_dispatch(
+        self,
+        n: Notification,
+        log: NotificationLoggerAdapter,
+        body: dict,
+        headers: Dict[bytes, List[str]],
+        pushkeys: List[str],
+        span: Span,
+    ) -> Optional[Tuple[list, list]]:
```
reivilibre, line 215:
```suggestion
    ) -> Optional[Tuple[List[str], List[str]]]:
```

sygnal/gcmpushkin.py
```diff
@@ -297,7 +319,9 @@ async def _request_dispatch(self, n, log, body, headers, pushkeys, span):
                 f"Unknown GCM response code {response.code}"
             )
 
-    async def _dispatch_notification_unlimited(self, n, device, context):
+    async def _dispatch_notification_unlimited(
+        self, n: Notification, device: Device, context: NotificationContext
+    ) -> Union[int, list]:
```
reivilibre, line 324:
```suggestion
    ) -> List[str]:
```
I don't think this can return an int — let me know if you spot something different

sygnal/gcmpushkin.py
```diff
@@ -325,7 +349,7 @@ async def _dispatch_notification_unlimited(self, n, device, context):
             }
 
             # TODO: Implement collapse_key to queue only one message per room.
-            failed = []
+            failed: list = []
```
reivilibre, line 352:
```suggestion
            failed: List[str] = []
```
iirc this is a list of pushkeys which are strings :)

sygnal/gcmpushkin.py
```diff
@@ -379,12 +403,12 @@ async def _dispatch_notification_unlimited(self, n, device, context):
             return failed
 
     @staticmethod
-    def _build_data(n, device):
+    def _build_data(n: Notification, device: Device) -> Dict:
```
reivilibre, line 406:
```suggestion
    def _build_data(n: Notification, device: Device) -> Dict[str, Any]:
```
(JSON-compatible dicts always have string keys, so this is usually a good enough type signature for us)

sygnal/notifications.py
```diff
@@ -14,32 +14,33 @@
...
 
 class Device:
     def __init__(self, raw):
-        self.app_id = None
-        self.pushkey = None
-        self.pushkey_ts = 0
-        self.data = None
-        self.tweaks = None
+        self.app_id: Optional[str] = None
+        self.pushkey: Optional[str] = None
```
reivilibre, line 40: These two are kind of fake optional.

Reckon you could do the dance to make it so that they're not annotated as optional? It probably needs to be something like:

```
// after if "app_id" not in raw: .... raise ...
raw_app_id = raw["app_id"]
if not isinstance(raw_app_id, str):
    raise InvalidNotificationException("Device with non-string app_id")
self.app_id: str = raw_app_id
```

(please do similarly for the pushkey, if you don't mind)

This just makes your life easier around the rest of the code that otherwise has to deal with this hypothetical (but impossible) case that these fields are `None`

sygnal/notifications.py
```diff
@@ -14,32 +14,33 @@
...
+        self.pushkey_ts: float = 0
```
reivilibre, line 41: do we know if this is float or int?
OK fine I'll stop being lazy and I'll look it up...

https://spec.matrix.org/unstable/push-gateway-api/#post_matrixpushv1notify
 says integer :)

(would you be able to validate that the number we pull out of the raw dict is an int before you shove it in this field?)

I think this kind of dance will do it, making use of `.get`'s defaults...:
```python
raw_pushkey_ts = raw.get("pushkey_ts", 0)
if not isinstance(raw_pushkey_ts, int):
    raise InvalidNotificationException("Device with non-integral pushkey_ts")
self.pushkey_ts: int = raw_pushkey_ts
```

sygnal/notifications.py
```diff
@@ -100,19 +101,19 @@ def __init__(self, name: str, sygnal: "Sygnal", config: Dict[str, Any]):
...
-    def get_config(self, key: str, default=None):
+    def get_config(self, key: str, default=None) -> Optional[Union[str, int, Dict]]:
         if key not in self.cfg:
             return default
         return self.cfg[key]
 
     async def dispatch_notification(
         self, n: Notification, device: Device, context: "NotificationContext"
-    ) -> List[str]:
+    ) -> Optional[List[str]]:
```
reivilibre, line 111: I think this should be backed out; the list of pushkeys should always be returned.

Probably replace `pass` below with `raise NotImplementedError()` as I guess Mypy will be sad that you're not returning anything

sygnal/notifications.py
```diff
@@ -167,7 +168,7 @@ def __init__(self, name: str, sygnal: "Sygnal", config: Dict[str, Any]):
 
     async def dispatch_notification(
         self, n: Notification, device: Device, context: "NotificationContext"
-    ) -> List[str]:
```
reivilibre, line 170: why are these annotations going away? :sob: 
I think they probably want to come back :)

reivilibre (changes requested):
Thanks for tackling this!
I think you might be noticing that trying to bite a lot at once will pull you into everywhere else in the code because everything's linked together.
A couple of places where you've used `cast` I've tried to replace with actual checks — `cast` should be thought of as 'oi mypy, you're too stupid to see that I'm right so just trust me unconditionally' — in general, we try to avoid it.

sygnal/notifications.py
reivilibre: Or maybe better:
I didn't know about this back in the day when I did the Sygnal → Python 3 port, but we could properly mark `Pushkin` as an Abstract Base Class (unhelpfully shortened to `ABC`).
You can then annotate  `@abstractmethod` on this method https://docs.python.org/3/library/abc.html#abc.abstractmethod
and replace the body with `...` (that's right: `...` is a valid expression in Python (crazy eh? I thought so at least!))

`class Pushkin:` → `class Pushkin(abc.ABC):`

sygnal/apnspushkin.py
```diff
@@ -38,9 +38,18 @@
...
+from sygnal.notifications import (
+    ConcurrencyLimitedPushkin,
+    Device,
+    Notification,
+    NotificationContext,
+)
 from sygnal.utils import NotificationLoggerAdapter, twisted_sleep
 
+if TYPE_CHECKING:
+    from .sygnal import Sygnal
```
squahtx, line 50: Could we stick to absolute imports for consistency? I'm aware this may have been copied from existing (inconsistent) code.

```suggestion
    from sygnal.sygnal import Sygnal
```

I'm about to put up a PR to convert all the existing relative imports into absolute imports. It might create conflicts with this PR unfortunately.

sygnal/apnspushkin.py
```diff
@@ -328,18 +352,21 @@ def _get_payload_event_id_only(self, n, device):
...
             n: The notification to construct a payload for.
-            device (Device): Device information to which the constructed payload
+            device: Device information to which the constructed payload
             will be sent.
             log: A logger.
 
         Returns:
             The APNs payload as nested dicts.
         """
+        assert n.sender is not None
```
squahtx, line 369: Unfortunately `sender` and `sender_display_name` are optional in the spec, so the existing code needs updating to handle the case where they're both missing.

https://spec.matrix.org/unstable/push-gateway-api/#post_matrixpushv1notify

sygnal/apnspushkin.py
```diff
@@ -471,7 +498,7 @@ def _get_payload_full(self, n, device, log):
 
         return payload
 
-    async def _send_notification(self, request):
+    async def _send_notification(self, request: NotificationRequest):
```
squahtx, line 501:
```py
from aioapns.common import NotificationResult
```
```suggestion
    async def _send_notification(self, request: NotificationRequest) -> NotificationResult:
```

sygnal/apnspushkin.py
squahtx: to elaborate, `list` really means `list[Any]` to mypy. There's an implicit hidden `Any` in there.

sygnal/gcmpushkin.py
```diff
@@ -34,7 +36,15 @@
...
-from .notifications import ConcurrencyLimitedPushkin
+from .notifications import (
+    ConcurrencyLimitedPushkin,
+    Device,
+    Notification,
+    NotificationContext,
+)
+
+if TYPE_CHECKING:
+    from .sygnal import Sygnal
```
squahtx, line 47:
```suggestion
    from sygnal.sygnal import Sygnal
```

sygnal/gcmpushkin.py
squahtx: I don't think this is `Optional`?

sygnal/apnspushkin.py
```diff
@@ -151,7 +160,7 @@ def __init__(self, name, sygnal, config):
                 loop=loop,
             )
 
-            self._report_certificate_expiration(certfile)
+            self._report_certificate_expiration(cast(str, certfile))
```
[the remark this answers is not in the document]
squahtx: We _could_ try something fancy with `get_config` so that it does the validation and returns the correct type at the same time?

```py
def get_config(key: str, type_: Type[T], default: DefaultT = None) -> Union[T, DefaultT]

...

certfile = self.get_config("certfile", str)
```

sygnal/notifications.py
squahtx: We can simplify these constructors for `Tweaks`, `Device` and `Counts` by using `dict.get`:
```py
    def __init__(self, raw: Dict[str, Any]):
         if "app_id" not in raw:
             raise InvalidNotificationException("Device with no app_id")
         if "pushkey" not in raw:
             raise InvalidNotificationException("Device with no pushkey")

       self.app_id: str = raw["app_id"]
       self.pushkey: str = raw["pushkey"]
       self.pushkey_ts: int = raw.get("pushkey_ts", 0)
       self.data: Optional[Dict[str, Any]] = raw.get("data")
       self.tweaks = Tweaks(raw.get("tweaks", {}))
```
(or similar)

(nb: `tweaks` is not `Optional`)

and if we wanted validation, we could define a helper function similar to my suggestion for `get_config`

sygnal/gcmpushkin.py
```diff
@@ -152,13 +164,15 @@ async def create(cls, name, sygnal, config):
         """
         return cls(name, sygnal, config)
 
-    async def _perform_http_request(self, body, headers):
+    async def _perform_http_request(
+        self, body: Dict, headers: Dict[bytes, List[str]]
+    ) -> Optional[Tuple[IResponse, str]]:
```
squahtx, line 169:
```suggestion
    ) -> Tuple[IResponse, str]:
```
I don't think this can return `None`?

squahtx:
For the `tuple index out of range` complaints, we might want to define
`Choppable = Union[Tuple[Literal["alert", "alert.body"]], Tuple[Literal["alert.loc-args"], int]]`
(ie. `Choppable` is either `("alert",)`, `("alert.body",)` or `("alert.loc-args", int)`)
then mypy should be clever enough to figure things out.

sygnal/apnspushkin.py
H-Shay: So the reasoning behind this was: `shaved_payload` is assigned to the return value of `apnstruncate.truncate` a few lines above this assertion, and the return value of `apnstruncate.truncate` is currently `Optional[Dict[str, Any]`, so mypy got upset that we were potentially passing a `None` into `_dispatch_request`. The mypy docs said to add an `assert foo is not none` in those types of cases, so that is what I did, although it seems like I should have handled it in another way.

H-Shay: And this is all because I was confused about the return type of `apnstruncate.truncate` which is wrong.

sygnal/apnspushkin.py
H-Shay: In the case where we have neither `sender` nor `sender_display_name` would it make sense to just set `from_display` to an empty string, or is there a more graceful way to handle that?

squahtx: There don't seem to be any good options for this case. Whatever we pick will end up spliced into one of these notification strings which may be in any language: https://github.com/vector-im/element-ios/blob/develop/Riot/Assets/en.lproj/Localizable.strings . Anything will do I guess.

sygnal/notifications.py
reivilibre: We definitely want validation.

```
         if "app_id" not in raw:
             raise InvalidNotificationException("Device with no app_id")
       self.app_id: str = raw["app_id"]
```
is a worst-of-both worlds approach where you get neither validation nor type errors (since `Any` will unify with `str` and everything pretends its fine)

sygnal/apnspushkin.py
reivilibre: ^ further on that, the spelling is `List` with a capital `L`. I'm sure there's some strange historical reason at play :/

DMRobertson: I wouldn't say it's that strange. `list[...]` wasn't legitimate syntax until Python 3.9, because the `list` object didn't have a `__getitem__` defined. That changed with https://www.python.org/dev/peps/pep-0585/ . The PEP suggests that you can get this in 3.7 with `from __future__ import annotations`. In any case, we support 3.6 so must continue to use the old syntax. (We can get away with using the newer syntax in stub files, as long as mypy is sufficiently new.)

sygnal/gcmpushkin.py
DMRobertson: > so this is usually a good enough type signature for us

Yes and no. The `Any` values really limit what the typechecker can spot and interrogate!

sygnal/notifications.py
H-Shay: I added a helper method to do some validation, which is quite similar to the `get_config` helper function, which makes me wonder if it makes sense to combine them? It seems that you would lose the specificity of the error message in `get_config`, but I don't know how to weigh that against having somewhat repeated code in a file. Thoughts?

sygnal/apnspushkin.py
```diff
@@ -109,7 +110,7 @@ def __init__(self, name: str, sygnal: "Sygnal", config: Dict[str, Any]) -> None:
                 nonunderstood,
             )
 
-        platform = self.get_config("platform", str)
+        platform = get_key(self.cfg, "platform", str)
```
squahtx, line 113: Was this intentional?

H-Shay: Nope, totally forgot to remove that. I was considering refactoring, see my question here: https://github.com/matrix-org/sygnal/pull/264#discussion_r745050336, then thought I should probably ask before diving into that and then obviously forgot to remove that line.

sygnal/gcmpushkin.py
```diff
@@ -212,7 +213,7 @@ async def _request_dispatch(
         headers: Dict[bytes, List[str]],
         pushkeys: List[str],
         span: Span,
-    ) -> Optional[Tuple[list, list]]:
+    ) -> Tuple[list, list]:
```
reivilibre, line 216:
```suggestion
    ) -> Tuple[List[str], List[str]]:
```
I think

sygnal/apnspushkin.py
```diff
@@ -328,21 +353,26 @@ def _get_payload_event_id_only(self, n, device):
...
             The APNs payload as nested dicts.
         """
-        from_display = n.sender
-        if n.sender_display_name is not None:
+        if not n.sender and not n.sender_display_name:
+            from_display = " "
+        elif n.sender_display_name is not None:
             from_display = n.sender_display_name
+        elif n.sender is not None:
+            from_display = n.sender
```
reivilibre, line 375: noting change in behaviour, though the spec says `sender` is optional so happy

reivilibre (changes requested):
there are merge conflicts, otherwise very close to merge I think

sygnal/apnspushkin.py
```diff
@@ -313,15 +313,11 @@ async def _dispatch_notification_unlimited(
                     span_parent.log_kv(
                         {"event": "temporary_fail", "retrying_in": retry_delay}
                     )
+                    await twisted_sleep(
+                        retry_delay, twisted_reactor=self.sygnal.reactor
+                    )
```
reivilibre, line 318:
```suggestion
                    if retry_number < self.MAX_TRIES - 1:
                        await twisted_sleep(
                            retry_delay, twisted_reactor=self.sygnal.reactor
                        )
```
subtle, but this stops it from sleeping when we've already given up.

(`try sleep try sleep try fail` rather than `try sleep try sleep try sleep fail`)

reivilibre (dismissed):
lgtm otherwise

reivilibre (approved):
Thank you!

PR #265, opened 2021-11-05
Use absolute imports for consistency

squahtx:


DMRobertson (approved):
I'm keen for consistency, and neutral on whether we should prefer relative or absolute imports.

Is there a linter check we can use to enforce this?

richvdh:
> I'm keen for consistency, and neutral on whether we should prefer relative or absolute imports.

I thought this was explicitly covered by https://github.com/matrix-org/synapse/blob/develop/docs/code_style.md, but it seems I thought wrong.

In the absence of our own rules, we generally defer to Google's style guide, which is generally comprehensive and well-reasoned. In this case it [says](https://google.github.io/styleguide/pyguide.html#23-packages):

> Import each module using the full pathname location of the module.

richvdh:
(and [also](https://google.github.io/styleguide/pyguide.html#224-decision):

> Do not use relative names in imports. 

)

squahtx:
> Is there a linter check we can use to enforce this?

A quick search found https://github.com/MarcoGorelli/absolufy-imports which will reformat your imports, and https://pypi.org/project/flake8-absolute-import/

PR #269, opened 2021-11-05
Fix overly generous matching of pushkin app IDs, introduced in 0.5.0

squahtx:
Prevent dots (.) from being interpreted as wildcards.
Only use pushkins which match the entire app ID, rather than a
substring.


sygnal/http.py
```diff
@@ -218,13 +218,13 @@ def find_pushkins(self, appid):
...
-        for key, value in self.sygnal.pushkins.items():
-            # The ".+" symbol is used in place of "*" symbol
-            regex = key.replace("*", ".+")
-            if re.search(regex, appid):
-                result.append(value)
-        return result
+        # otherwise, find any pushkins whose appid patterns match
+        pushkins = []
+        for pushkin_appid, pushkin in self.sygnal.pushkins.items():
+            pattern = glob_to_regex(pushkin_appid, ignore_case=False)
```
reivilibre, line 224: It seems like it would be nice to precompile these, but it's less easy than the other case so I guess it depends on how ugly you think that'd be. (It also short-circuits to doing a direct check in the dict for the exact app ID, so I don't know how commonly this case is hit.)

squahtx: I've changed the patterns to be explicitly cached on startup.

reivilibre (dismissed):
Mypy's sad, otherwise looks good.

squahtx:
> Mypy's sad, otherwise looks good.

It'll be because it needs #268 to be merged first.

squahtx:
Nothing's really changed since the last review. It's only a merge from main followed by a switch to `matrix-python-common`.

reivilibre (approved):

PR #270, opened 2021-11-05
Fix `GcmPushkin` to handle app ID globs correctly, introduced in 0.5.0

squahtx:
Addresses #255 together with ~~#268~~ #281 and #269.

### DIFF
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -27,14 +27,18 @@ from twisted.web.http_headers import Headers
 
 from sygnal.exceptions import (
     NotificationDispatchException,
+    PushkinSetupException,
     TemporaryNotificationDispatchException,
 )
 from sygnal.helper.context_factory import ClientTLSOptionsFactory
 from sygnal.helper.proxy.proxyagent_twisted import ProxyAgent
-from sygnal.utils import NotificationLoggerAdapter, json_decoder, twisted_sleep
-
-from .exceptions import PushkinSetupException
-from .notifications import ConcurrencyLimitedPushkin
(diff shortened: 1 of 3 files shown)


sygnal/gcmpushkin.py
```diff
@@ -300,13 +304,22 @@ async def _request_dispatch(self, n, log, body, headers, pushkeys, span):
     async def _dispatch_notification_unlimited(self, n, device, context):
         log = NotificationLoggerAdapter(logger, {"request_id": context.request_id})
 
+        # `_dispatch_notification_unlimited` gets called once for each device in the
+        # `Notification` with a matching app ID. We do something a little dirty and
+        # perform all of our dispatches the first time we get called for a
+        # `Notification` and do nothing for the rest of the times we get called.
+        app_id_pattern = glob_to_regex(self.name, ignore_case=False)
```
reivilibre, line 311: Please can you pre-compile this at startup? `re.compile` ~~is~~ was(?)* expensive.

*I believe it used to bypass the cache. Looking at the code in my Python 3.9 install, it seems like it uses the cache now.
However, pre-compiling is probably still best practice, especially since it's trivial to do in this case.

squahtx: I've changed it to be explicitly cached on startup.

reivilibre (dismissed):
Mypy's sad here too, otherwise looks good

squahtx:
> Mypy's sad here too, otherwise looks good

It'll be because it needs #268 to be merged first.

squahtx:
Nothing's really changed since the last review. It's only a merge from main followed by a switch to `matrix-python-common`.

reivilibre (approved):

PR #271, opened 2021-11-09
Add type hints to webpushpushkin.py

H-Shay:
Hopefully this is a little easier to parse. Just one file. There are a few mypy errors but I believe they will be resolved once the type hints for `notifications.py` are accepted and merged.

### DIFF
--- a/sygnal/webpushpushkin.py
+++ b/sygnal/webpushpushkin.py
@@ -241,9 +241,10 @@ class WebpushPushkin(ConcurrencyLimitedPushkin):
         """
         payload = {}
 
-        default_payload = device.data.get("default_payload")
-        if isinstance(default_payload, dict):
-            payload.update(default_payload)
+        if device.data:
+            default_payload = device.data.get("default_payload")
+            if isinstance(default_payload, dict):
+                payload.update(default_payload)
 
         for attr in [
             "room_id",
--- a/sygnal/webpushpushkin.py
+++ b/sygnal/webpushpushkin.py
@@ -22,7 +22,6 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Pattern
 from urllib.parse import urlparse
 
 import py_vapid
-import requests
 from prometheus_client import Gauge, Histogram
 from py_vapid import Vapid, VapidException
 from pywebpush import webpush
@@ -406,7 +405,9 @@ class HttpDelayedRequest:
         self.data = data
         self.vapid_headers = vapid_headers
 
-    def execute(self, http_agent: ProxyAgent, low_priority: bool, topic: bytes) -> IResponse:
+    def execute(
+        self, http_agent: ProxyAgent, low_priority: bool, topic: bytes
+    ) -> IResponse:
         body_producer = FileBodyProducer(BytesIO(self.data))
         # Convert the headers to the camelcase version.
         headers = {
--- a/sygnal/webpushpushkin.py
+++ b/sygnal/webpushpushkin.py
@@ -21,10 +21,9 @@ from io import BytesIO
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Pattern
 from urllib.parse import urlparse
 
-import py_vapid
 from prometheus_client import Gauge, Histogram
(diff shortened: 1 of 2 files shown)


sygnal/webpushpushkin.py
```diff
@@ -18,22 +18,32 @@
 from base64 import urlsafe_b64encode
 from hashlib import blake2s
 from io import BytesIO
-from typing import List, Optional, Pattern
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Pattern
 from urllib.parse import urlparse
 
+import py_vapid
 from prometheus_client import Gauge, Histogram
 from py_vapid import Vapid, VapidException
```
DMRobertson, line 26: Nit: it would be nicer to import `py_vapid.CaseSensitiveDict` on line 26 of the RHS here, along with the other imports from that module.

DMRobertson (changes requested):
Mypy appears to be unhappy:

```
 sygnal/webpushpushkin.py:150: error: List item 0 has incompatible type "Optional[Any]"; expected "str"  [list-item]
sygnal/webpushpushkin.py:180: error: List item 0 has incompatible type "Optional[Any]"; expected "str"  [list-item]
sygnal/webpushpushkin.py:229: error: Argument 3 to "_handle_response" of "WebpushPushkin" has incompatible type "Optional[Any]"; expected "str"  [arg-type]
sygnal/webpushpushkin.py:232: error: List item 0 has incompatible type "Optional[Any]"; expected "str"  [list-item]
sygnal/webpushpushkin.py:250: error: Item "None" of "Optional[Any]" has no attribute "get"  [union-attr]
```

I think most of these would be fixed by marking `Device.pushkin: str`.
The last one seems to want an annotation for `Device.data`.

(Perhaps you have these locally in another commit?)

sygnal/webpushpushkin.py
```diff
@@ -75,7 +85,7 @@ class WebpushPushkin(ConcurrencyLimitedPushkin):
         "ttl",
     } | ConcurrencyLimitedPushkin.UNDERSTOOD_CONFIG_FIELDS
 
-    def __init__(self, name, sygnal, config):
+    def __init__(self, name: str, sygnal: Sygnal, config: Dict[str, Any]):
```
squahtx, line 88: `Sygnal` may need to be quoted since it's conditionally imported.
```suggestion
    def __init__(self, name: str, sygnal: "Sygnal", config: Dict[str, Any]):
```

H-Shay:
> Mypy appears to be unhappy:

So I mentioned this at the top of the PR that I think these errors will be resolved once the changes to `notifications.py` are merged-is the best thing to do to wait until that happens? Just wasn't sure how to manage the dependencies amongst mypy'd files in separate PR's.

DMRobertson:
> > Mypy appears to be unhappy:
> 
> So I mentioned this at the top of the PR that I think these errors will be resolved once the changes to `notifications.py` are merged-is the best thing to do to wait until that happens? Just wasn't sure how to manage the dependencies amongst mypy'd files in separate PR's.

Arg, sorry. You did indeed; I just dived straight into the diff. I think the changes you refer to are in https://github.com/matrix-org/sygnal/pull/264 ?

FWIW, I've found this workflow to be okayish for managing this kind of dependency:

I've handled this in the past by starting a second branch based on the first change's branch. Then I target the PR at `main/develop/master` and mark it as a draft.  That means that the PR includes both commits from the first and second branch, so I note in the description "depends on #<number>, first interesting commit is <hash>".

I typically end up
- adding to the first until review passes
- merging the first into main
- rebase the second onto main
- go through review for the second

DMRobertson (commented):
LGTM. Maybe @squahtx  wants to have their say too? Otherwise, I think this is good to go once #264 is merged.

squahtx:
> LGTM. Maybe @squahtx wants to have their say too? Otherwise, I think this is good to go once #264 is merged.

It all looks good to me!

(I wonder if it's tractable to change `Dict[str, Any]` to `Dict[str, object]` so that mypy makes us add validation. But that's something for another time.)

DMRobertson (approved):
Looking good!

PR #273, opened 2021-11-16
Add type hints to `sygnal/http.py`

H-Shay:
As the title states. Also want to acknowledge that I used DMRobertson's type stub work in Sydent in this PR, specifically `stubs/twisted/web/http.pyi`.

### DIFF
--- a/changelog.d/265.misc
+++ b/changelog.d/265.misc
@@ -1,0 +1,1 @@
+Use absolute imports for consistency.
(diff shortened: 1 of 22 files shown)


squahtx (approved):
LGTM!

I wonder if there's a way to share stubs between repos?

DMRobertson:
> I wonder if there's a way to share stubs between repos?

Git submodules would be my suggestion, assuming we don't want to publish a stubs package to pypi

reivilibre:
the answer there might be to upstream the type annotations and get rid of the stubs altogether :sweat_smile:

PR #275, opened 2021-11-16
Add type hints to `sygnal.py` 

H-Shay:
As the title says.

### DIFF
--- a/sygnal/sygnal.py
+++ b/sygnal/sygnal.py
@@ -20,7 +20,7 @@ import logging
 import logging.config
 import os
 import sys
-from typing import Any, Dict
+from typing import Any, Dict, Set
 
 import opentracing
 import prometheus_client
@@ -36,7 +36,7 @@ from sygnal.notifications import Pushkin
 
 logger = logging.getLogger(__name__)
 
-CONFIG_DEFAULTS: dict = {
+CONFIG_DEFAULTS: Dict[str, Any] = {
     "http": {"port": 5000, "bind_addresses": ["127.0.0.1"]},
     "log": {"setup": {}, "access": {"x_forwarded_for": False}},
     "metrics": {
@@ -70,7 +70,7 @@ class Sygnal(object):
         """
         self.config = config
         self.reactor = custom_reactor
-        self.pushkins: Dict[str, Pushkin] = {}
+        self.pushkins: Dict[str, Any] = {}
         self.tracer = tracer
 
         logging_dict_config = config["log"]["setup"]
@@ -172,7 +172,7 @@ class Sygnal(object):
         clarse = getattr(pushkin_module, to_construct)
         return await clarse.create(app_name, self, app_config)
 
-    async def make_pushkins_then_start(self):
+    async def make_pushkins_then_start(self) -> None:
         for app_id, app_cfg in self.config["apps"].items():
             try:
                 self.pushkins[app_id] = await self._make_pushkin(app_id, app_cfg)
@@ -195,7 +195,7 @@ class Sygnal(object):
             logger.info("Starting listening on %s port %d", interface, port)
             self.reactor.listenTCP(port, pushgateway_api.site, interface=interface)
 
-    def run(self):
+    def run(self) -> None:
         """
         Attempt to run Sygnal and then exit the application.
         """
@@ -246,7 +246,9 @@ def check_config(config: Dict[str, Any]) -> None:
     """
     UNDERSTOOD_CONFIG_FIELDS = CONFIG_DEFAULTS.keys()
 
-    def check_section(section_name, known_keys, cfgpart=config):
+    def check_section(
+        section_name: str, known_keys: Set[str], cfgpart: Dict[str, Any] = config
+    ) -> None:
         nonunderstood = set(cfgpart[section_name].keys()).difference(known_keys)
         if len(nonunderstood) > 0:
             logger.warning(
--- a/sygnal/sygnal.py
+++ b/sygnal/sygnal.py
@@ -20,6 +20,7 @@ import logging
 import logging.config
 import os
 import sys
+from typing import Any, Dict
 
 import opentracing
 import prometheus_client
@@ -31,6 +32,7 @@ from twisted.python import log as twisted_log
 from twisted.python.failure import Failure
 
 from sygnal.http import PushGatewayApiServer
+from sygnal.notifications import Pushkin
 
 logger = logging.getLogger(__name__)
 
@@ -53,17 +55,22 @@ CONFIG_DEFAULTS: dict = {
 
 
 class Sygnal(object):
-    def __init__(self, config, custom_reactor, tracer=opentracing.tracer):
+    def __init__(
+        self,
+        config: Dict[str, Any],
+        custom_reactor: asyncioreactor.AsyncioSelectorReactor,
+        tracer=opentracing.tracer,
+    ):
         """
         Object that holds state for the entirety of a Sygnal instance.
         Args:
-            config (dict): Configuration for this Sygnal
+            config: Configuration for this Sygnal
             custom_reactor: a Twisted Reactor to use.
             tracer (optional): an OpenTracing tracer. The default is the no-op tracer.
         """
         self.config = config
         self.reactor = custom_reactor
-        self.pushkins = {}
+        self.pushkins: Dict[str, Pushkin] = {}
         self.tracer = tracer
 
         logging_dict_config = config["log"]["setup"]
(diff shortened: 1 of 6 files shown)


sygnal/sygnal.py
```diff
@@ -52,18 +54,23 @@
 }
 
 
-class Sygnal:
-    def __init__(self, config, custom_reactor, tracer=opentracing.tracer):
+class Sygnal(object):
```
DMRobertson, line 57: Inheriting from object is redundant in Python 3.

```suggestion
class Sygnal:
```

It used to have an effect in Python 2.3--2.7, to opt into "new-style" rather than "old-style" classes. Details [here](https://docs.python.org/2/reference/datamodel.html#new-style-and-classic-classes) and [here](http://python-history.blogspot.com/2010/06/inside-story-on-new-style-classes.html) if you're really curious.

Everything is a new-style class in Python 3.

sygnal/sygnal.py
```diff
@@ -52,18 +54,23 @@
...
+    def __init__(
+        self,
+        config: Dict[str, Any],
+        custom_reactor,
+        tracer=opentracing.tracer,
```
DMRobertson, line 62: Are there any obvious annotations for these two we could use?

(Not sure if your goal is "pass mypy" or "pass mypy with `--disallow-untyped-defs`"?)

DMRobertson (changes requested):

sygnal/sygnal.py
```diff
@@ -52,18 +54,23 @@
...
         Args:
-            config (dict): Configuration for this Sygnal
+            config: Configuration for this Sygnal
             custom_reactor: a Twisted Reactor to use.
             tracer (optional): an OpenTracing tracer. The default is the no-op tracer.
         """
         self.config = config
         self.reactor = custom_reactor
-        self.pushkins = {}
+        self.pushkins: Dict[str, Any] = {}
```
squahtx, line 73:
```suggestion
        self.pushkins: Dict[str, Pushkin] = {}
```

sygnal/sygnal.py
H-Shay: So I struggled with finding an appropriate type for `custom_reactor`. At first I had `asyncioreactor.AsyncioSelectorReactor` which is what is passed in `main`, but this caused problems in `test_utils.py`, where we pass a `ExtendedMemoryReactorClock` instead. I looked in the Twisted docs but couldn't find a reactor that encompassed both these classes so I defaulted to putting nothing in this case. I am sure there is a better solution.

DMRobertson: You might be able to use one of Twisted's reactor interfaces, like [IReactorCore](https://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.IReactorCore.html). But I suspect that won't work because we probably use methods from more than one interface.

In Sydent I ended up doing this based on what we do in Synapse: https://github.com/matrix-org/sydent/blob/cf27fce07798c9efeea901ddee4211574840024d/sydent/sydent.py#L91-L106

We could try something similar here. Wanna give it a go?

sygnal/sygnal.py
```diff
@@ -20,21 +20,32 @@
 import logging.config
 import os
 import sys
+from typing import Any, Dict, Set, Union, cast
 
+import jaeger_client
```
DMRobertson, line 25: is `jaeger_client` an optional dependency?

Oh, [apparently not](https://github.com/matrix-org/sygnal/blob/d9fb1a3d9d45cdc42435b7a87d1a036471d5ef51/setup.py#L31). Never mind then.

sygnal/sygnal.py
```diff
@@ -52,18 +63,33 @@
...
+    pass
+
+
 class Sygnal:
-    def __init__(self, config, custom_reactor, tracer=opentracing.tracer):
+    def __init__(
+        self,
+        config: Dict[str, Any],
+        custom_reactor: SygnalReactor,
+        tracer: Union[Tracer, jaeger_client.Tracer] = opentracing.tracer,
```
DMRobertson, line 81: I think you should be able to just use `Tracer`, because [jaeger_client.Tracer inherits from opentracing.Tracer](https://github.com/jaegertracing/jaeger-client-python/blob/a81052e73d238692ed5c6101aab420f51d414a64/jaeger_client/tracer.py#L46).

tests/testutils.py
```diff
@@ -75,13 +75,14 @@ def setUp(self):
...
-        self.sygnal = Sygnal(config, reactor)
+        self.sygnal = Sygnal(config, reactor)  # type: ignore
         self.reactor = reactor
 
         start_deferred = ensureDeferred(self.sygnal.make_pushkins_then_start())
 
         while not start_deferred.called:
             # we need to advance until the pushkins have started up
+            assert isinstance(self.sygnal.reactor, ExtendedMemoryReactorClock)
             self.sygnal.reactor.advance(1)
```
DMRobertson, line 86: Since we already have the reactor, maybe use `self.reactor.advance` etc. here and in the two cases below. I think that will let us avoid the assertions.

(Nothing wrong with as it is, just trying to keep things concise where possible)

tests/test_http.py
```diff
@@ -62,7 +64,10 @@ def setUp(self):
 
         self.apns_pushkin_snotif = MagicMock()
         for key, value in self.sygnal.pushkins.items():
-            value._send_notification = self.apns_pushkin_snotif
+            assert isinstance(value, ApnsPushkin)
+            # type safety: ignore is used here due to mypy not handling monkeypatching,
+            # see https://github.com/python/mypy/issues/2427
+            value._send_notification = self.apns_pushkin_snotif  # type: ignore
```
DMRobertson, line 70: Where possible, prefer `#type: ignore[code]` where `code` is one of mypy's error codes. See https://mypy.readthedocs.io/en/stable/error_codes.html and the two pages following it for gory details.

(but easier to remove the comment and see what the error code you get in the terminal is.)

tests/test_gcm.py
```diff
@@ -183,13 +190,14 @@ def test_fcm_options(self):
...
-        gcm = self.sygnal.pushkins["com.example.gcm.ios"]
+        gcm = self.get_test_pushkin("com.example.gcm.ios")
         gcm.preload_with_response(
             200, {"results": [{"registration_id": "spqr_new", "message_id": "msg42"}]}
         )
 
         resp = self._request(self._make_dummy_notification([DEVICE_EXAMPLE_IOS]))
 
         self.assertEqual(resp, {"rejected": []})
+        assert gcm.last_request_body is not None
```
DMRobertson, line 201: I was going to say:

> I wonder if we can use [`self.assertIsNotNone`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsNotNone) here, so that the assertion is more formally part of the test?

But I don't think that mypy will recognise this as narrowing `gcm.last_request_body` to something that's not-None. So probably best to ignore this comment and leave it as it is.

sygnal/sygnal.py
```diff
@@ -271,14 +299,16 @@ def check_section(section_name, known_keys, cfgpart=config):
     check_section("sentry", {"enabled", "dsn"}, cfgpart=config["metrics"])
 
 
-def merge_left_with_defaults(defaults, loaded_config):
+def merge_left_with_defaults(
```
DMRobertson, line 302: I haven't looked very carefully, but 

```
merged = merge_left_with_defaults(defaults, loaded_config)
```

sounds to me like it would be the same as the following two lines:

```
merged = copy.deepcopy(loaded_config)
merged.update(defaults)
```

Let's not change this. Just mentioning it for interest!

sygnal/sygnal.py
```diff
@@ -186,9 +212,9 @@ async def make_pushkins_then_start(self):
         port = int(self.config["http"]["port"])
         for interface in self.config["http"]["bind_addresses"]:
             logger.info("Starting listening on %s port %d", interface, port)
-            self.reactor.listenTCP(port, pushgateway_api.site, interface=interface)
+            self.reactor.listenTCP(port, pushgateway_api.site, 50, interface=interface)
```
DMRobertson, line 215: Ahh yes, I remember this. PosixReactorBase provides a default value for this arg, even though IReactorTCP doesn't!

https://github.com/matrix-org/sydent/blob/7bdaac2774840e1f9697d30515f367c5c8c16866/sydent/http/httpserver.py#L148

DMRobertson (changes requested):
Looking good. I think the only two changes I'm particularly keen on are the Tracer and the `#type: ignore[code]` one. The rest are very minor or comments for interest.

DMRobertson (approved):

DMRobertson:
Thanks @H-Shay, looks great!

PR #276, opened 2021-11-22
Add type hints to `sygnal/utils.py`

H-Shay:
As the title states.

### DIFF
--- a/sygnal/utils.py
+++ b/sygnal/utils.py
@@ -15,11 +15,15 @@
 import json
 import re
 from logging import LoggerAdapter
+from typing import Any, MutableMapping, Pattern, Tuple, Union
 
+from twisted.internet import asyncioreactor
 from twisted.internet.defer import Deferred
 
 
-async def twisted_sleep(delay, twisted_reactor):
+async def twisted_sleep(
+    delay: float, twisted_reactor: asyncioreactor.AsyncioSelectorReactor
+) -> None:
     """
     Creates a Deferred which will fire in a set time.
     This allows you to `await` on it and have an async analogue to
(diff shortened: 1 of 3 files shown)


sygnal/utils.py
```diff
@@ -65,7 +71,7 @@ def glob_to_regex(glob):
     return re.compile(r"\A" + res + r"\Z", re.IGNORECASE)
 
 
-def _reject_invalid_json(val):
+def _reject_invalid_json(val: Any) -> None:
```
DMRobertson, line 74: Could also use `NoReturn` here, but I don't think there's any practical benefit to doing so.

sygnal/utils.py
```diff
@@ -37,17 +41,19 @@ async def twisted_sleep(delay, twisted_reactor):
...
 class NotificationLoggerAdapter(LoggerAdapter):
-    def process(self, msg, kwargs):
+    def process(
+        self, msg: str, kwargs: MutableMapping[str, Any]
+    ) -> Tuple[str, MutableMapping[str, Any]]:
         return f"[{self.extra['request_id']}] {msg}", kwargs
 
 
-def glob_to_regex(glob):
+def glob_to_regex(glob: str) -> Union[Pattern, Pattern[str]]:
```
DMRobertson, line 50: I don't understand the return type here---why can't we just say it returns a `Pattern[str]` in all cases?

sygnal/utils.py
```diff
@@ -15,11 +15,15 @@
...
+from typing import TYPE_CHECKING, Any, MutableMapping, Pattern, Tuple, Union
 
 from twisted.internet.defer import Deferred
 
+if TYPE_CHECKING:
+    from sygnal.sygnal import SygnalReactor
 
-async def twisted_sleep(delay, twisted_reactor):
+
+async def twisted_sleep(delay: float, twisted_reactor: "SygnalReactor") -> None:
```
DMRobertson, line 26: This is fine as it is, but I wanted to mention a couple of alternatives here. One be to write `twisted_reactor: IReactorTime` and only import `IReactorTime`; a second would be to move `SygnalReactor` into a new file, say `sygnal/types.py`.

sygnal/utils.py
H-Shay: I nicked this from somewhere and now I can't remember where, but I think you are correct we can use the narrower case. I will change this.

sygnal/utils.py
H-Shay: Great, I appreciate knowing alternative solutions, it really helps to have more tools in my arsenal.

DMRobertson (approved):
I'm happy if mypy is!

PR #278, opened 2021-11-29
Convert the README to use Markdown rather than reStructuredText for consistency and familiarity.

reivilibre:
I personally get tripped up by rST's not-quite-Markdown false friends and I also reckon that Markdown is more widely-known.

Rendered: https://github.com/matrix-org/sygnal/blob/05b2202efae4ae02994a65cc96910b66734c1898/README.md


clokep:
> (This is also better for consistency with the rest of our projects! :))

Both synapse and sydent seem to use rST?

setup.py
```diff
@@ -74,5 +74,5 @@ def read_file(path_segments: Sequence[str]) -> str:
         setup_requires=["setuptools_scm"],
         install_requires=INSTALL_REQUIRES,
         extras_require=EXTRAS_REQUIRE,
-        long_description=read_file(("README.rst",)),
+        long_description=read_file(("README.md",)),
```
clokep, line 77: I think you would want to include what the content type of this file is. (I think it is `long_description_content_type`?)

reivilibre:
> Both synapse and sydent seem to use rST?

Gah. I don't know what planet I was on imagining that they don't...

DMRobertson (approved):
I have no strong opinions on Markdown versus RST, but this looks like a sensible conversion from one to the other.

(For completeness, I had a [good experience](https://thompsons-v.readthedocs.io/en/master/index.html) with Sphinx + autodoc + doctest + intersphinx + viewcode, so have some lingering positive emotions attributed to RST)

PR #280, opened 2021-12-02
Treat more APNs errors as permanent rejections

### LINKED ISSUE #279: We should treat more kinds of APNs errors as permanent rejections.
See https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/handling_notification_responses_from_apns for a list of error codes.

We currently seem to only treat `410 Unregistered` as an error that leads to a rejection making its way back to the homeserver.

However, perhaps we should also reject on `DeviceTokenNotForTopic` and some more (which we need to decide).

The consideration here is that we should only reject when an error is not Sygnal's fault and is permanent (since the homeserver will delete that pushkey once it's rejected).

squahtx:
Fixes #279

### DIFF
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -83,13 +83,16 @@ class ApnsPushkin(ConcurrencyLimitedPushkin):
     # See https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/handling_notification_responses_from_apns  # noqa: E501
     # for the full list of possible errors.
     TOKEN_ERRORS = {
-        (400, "BadDeviceToken"),  # Invalid token
+        # A client has uploaded an invalid token.
+        (400, "BadDeviceToken"),
         # `DeviceTokenNotForTopic` may be due to a bad token or a Sygnal
-        # misconfiguration. It's alright for the homeserver to forget a valid token,
-        # since iOS clients are expected to re-upload their device tokens periodically.
+        # misconfiguration. In the event of a misconfiguration, clients will need to
+        # reupload their tokens to their homeserver. Element iOS and clients based on
+        # matrix-ios-kit already do this periodically.
         (400, "DeviceTokenNotForTopic"),
         (400, "TopicDisallowed"),
-        (410, "Unregistered"),  # Deactivated token
+        # The token is no longer valid, probably because the app has been uninstalled.
+        (410, "Unregistered"),
     }
 
     MAX_TRIES = 3
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -88,8 +88,7 @@ class ApnsPushkin(ConcurrencyLimitedPushkin):
         # `DeviceTokenNotForTopic` may be due to a token for a different app or an
         # incorrect topic in the Sygnal configuration. In the event of a
         # misconfiguration, clients will need to reupload their tokens to their
-        # homeserver. Element iOS and clients based on matrix-ios-kit already do this
-        # periodically.
+        # homeserver.
         (400, "DeviceTokenNotForTopic"),
         (400, "TopicDisallowed"),
         # The token is no longer valid, probably because the app has been uninstalled.
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -80,7 +80,7 @@ class ApnsPushkin(ConcurrencyLimitedPushkin):
     """
 
     # Errors for which the token should be rejected and not reused
-    # See https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/handling_notification_responses_from_apns
+    # See https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/handling_notification_responses_from_apns  # noqa: E501
     # for the full list of possible errors.
     TOKEN_ERRORS = {
         (400, "BadDeviceToken"),  # Invalid token
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -79,9 +79,18 @@ class ApnsPushkin(ConcurrencyLimitedPushkin):
     Relays notifications to the Apple Push Notification Service.
     """
 
-    # Errors for which the token should be rejected
-    TOKEN_ERROR_REASON = "Unregistered"
-    TOKEN_ERROR_CODE = 410
+    # Errors for which the token should be rejected and not reused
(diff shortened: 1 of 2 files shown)


sygnal/apnspushkin.py
```diff
@@ -79,9 +79,18 @@ class ApnsPushkin(ConcurrencyLimitedPushkin):
...
-    TOKEN_ERROR_REASON = "Unregistered"
-    TOKEN_ERROR_CODE = 410
+    # Errors for which the token should be rejected and not reused
+    # See https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/handling_notification_responses_from_apns
+    # for the full list of possible errors.
+    TOKEN_ERRORS = {
+        (400, "BadDeviceToken"),  # Invalid token
+        # `DeviceTokenNotForTopic` may be due to a bad token or a Sygnal
+        # misconfiguration. It's alright for the homeserver to forget a valid token,
+        # since iOS clients are expected to re-upload their device tokens periodically.
```
squahtx, line 89: In particular, Element iOS re-uploads its push token every time it's foregrounded (along with syncing and all the other stuff it does).

https://github.com/matrix-org/matrix-ios-kit/blob/b85b736313bec0592bd1cabc68035d97f5331137/MatrixKit/Models/Account/MXKAccount.m#L1152

sygnal/apnspushkin.py
```diff
@@ -239,10 +248,7 @@ async def _dispatch_request(
         else:
             # .description corresponds to the 'reason' response field
             span.set_tag("apns_reason", response.description)
-            if (
-                code == self.TOKEN_ERROR_CODE
-                or response.description == self.TOKEN_ERROR_REASON
-            ):
+            if (code, response.description) in self.TOKEN_ERRORS:
```
squahtx, line 251: There's an existing test that covers the `(410, "Unregistered")` behaviour.

sygnal/apnspushkin.py
```diff
@@ -79,9 +79,18 @@ class ApnsPushkin(ConcurrencyLimitedPushkin):
...
     """
 
-    # Errors for which the token should be rejected
-    TOKEN_ERROR_REASON = "Unregistered"
-    TOKEN_ERROR_CODE = 410
+    # Errors for which the token should be rejected and not reused
+    # See https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/handling_notification_responses_from_apns  # noqa: E501
+    # for the full list of possible errors.
+    TOKEN_ERRORS = {
+        (400, "BadDeviceToken"),  # Invalid token
```
clokep, line 86: This would mean that the client gave us a bogus token?

sygnal/apnspushkin.py
clokep: > since iOS clients are expected to re-upload their device tokens periodically

Is this just part of how APNS works or is it specific to Element iOS?

sygnal/apnspushkin.py
```diff
@@ -79,9 +79,18 @@ class ApnsPushkin(ConcurrencyLimitedPushkin):
...
+        # `DeviceTokenNotForTopic` may be due to a bad token or a Sygnal
+        # misconfiguration. It's alright for the homeserver to forget a valid token,
+        # since iOS clients are expected to re-upload their device tokens periodically.
+        (400, "DeviceTokenNotForTopic"),
+        (400, "TopicDisallowed"),
```
clokep, line 91: Do both of these mean a broken client?

sygnal/apnspushkin.py
clokep: Previously we would error on `410` or `Unregistered`, but from their docs these should always be grouped so this shold be OK. 👍 

Is it worth logging or anything that we're deactivating this?

squahtx: Good spot, I actually misread the condition as an `and`, since I wasn't expecting an `or`!

Some digging reveals that we're already logging the tokens we reject here:
https://github.com/matrix-org/sygnal/blob/main/sygnal/http.py#L283-L287

sygnal/apnspushkin.py
squahtx: Bad wording on my part here. It's specific to Element iOS and anything else built on matrix-ios-kit.

At worst, users would need to log out and back in again once the config is fixed. Though if you're in control of the Sygnal instance, you're presumably also responsible for the associated app and can update it to re-upload tokens.

sygnal/apnspushkin.py
squahtx: Yep

sygnal/apnspushkin.py
squahtx: I think both of these mean a developer has messed up in some way. I'm not 100% clear on the how. I think the former means the token is valid but for a different app ID and the latter seems to be something to do with mixing up VOIP and non-VOIP tokens/topics.

sygnal/apnspushkin.py
```diff
@@ -79,9 +79,21 @@ class ApnsPushkin(ConcurrencyLimitedPushkin):
...
+    # Errors for which the token should be rejected and not reused
+    # See https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/handling_notification_responses_from_apns  # noqa: E501
+    # for the full list of possible errors.
+    TOKEN_ERRORS = {
+        # A client has uploaded an invalid token.
+        (400, "BadDeviceToken"),
+        # `DeviceTokenNotForTopic` may be due to a bad token or a Sygnal
+        # misconfiguration. In the event of a misconfiguration, clients will need to
+        # reupload their tokens to their homeserver. Element iOS and clients based on
+        # matrix-ios-kit already do this periodically.
```
clokep, line 91: Sygnal is supposed to be agnostic to Element iOS, but this seems to give good guidance here or what's going on.

I think matrix-ios-kit is also being deprecated?

```suggestion
        # reupload their tokens to their homeserver.
```

sygnal/apnspushkin.py
clokep: Alright! Thanks for the explanation!

sygnal/apnspushkin.py
clokep: That log-line is really confusing it says it was successful, but there were rejections.

I guess my question is -- how is someone supposed to debug *why* their request was rejected? Or does that not matter?

clokep (approved):
I think this looks good now! 👍

squahtx:
Thanks for reviewing!

PR #281, opened 2021-12-03
Move `glob_to_regex` to `matrix-python-common`

squahtx:
Replaces #268

Part of the work to address #255

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ INSTALL_REQUIRES = [
     "importlib_metadata",
     "pywebpush>=1.13.0",
     "py-vapid>=1.7.0",
+    "matrix-common==1.0.0",
 ]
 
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -36,6 +36,7 @@
     "importlib_metadata",
     "pywebpush>=1.13.0",
     "py-vapid>=1.7.0",
+    "matrix-common==1.0.0",
```
clokep, line 39: I don't remember if we had a conversation, but 👍 on pinning this. I think we want exact versions here since we're just trying to avoid vendoring this code.

clokep (approved):
Seems reasonable to me!

setup.py
squahtx: I seem to remember someone telling me to pin the `matrix-common` import, but can't remember where that was.

PR #282, opened 2021-12-04
Add type stubs for opentracing

H-Shay:
When looking at how to get Sygnal to 80% precision, I noticed many of the imprecise lines were either from intentional `Any`s in our code (i.e. Json dicts with `Dict[str, Any]` as the type and similar situations) or un-typed third-party code, so I tackled stubbing some of those libraries. More to come.


sygnal/apnspushkin.py
```diff
@@ -230,7 +230,7 @@ async def _dispatch_request(
 
         code = int(response.status)
 
-        span.set_tag(tags.HTTP_STATUS_CODE, code)
+        span.set_tag(str(tags.HTTP_STATUS_CODE), code)
```
squahtx, line 233: I'm not sure why but on my install `HTTP_STATUS_CODE` is a string:
```
>>> import opentracing.tags
>>> opentracing.tags.HTTP_STATUS_CODE
'http.status_code'
```
I'm on opentracing 2.4.0 for reference

reivilibre: > I'm not sure why but on my install HTTP_STATUS_CODE is a string

With context and thought, this sounds perfectly reasonable to me — it's a standardised name (key) for the tag where you put the status code of an HTTP request.

reivilibre:
Sorry to be the bearer of bad news (though this might save you the hassle of maintaining these later, so maybe it's good news!)

There appears to be a `types-opentracing` packages (this is a common theme; worth checking for `types-xxx` and `xxx-stubs` when wanting stubs for a module `xxx`).

If you remove the lines from `mypy.ini` that skip opentracing and do `pip install types-opentracing`, does it lead to great success?

If so, that may be the way to go — I suppose it would need to be added as a dev dependency.

reivilibre (changes requested):
will mark this as requested changes until we establish whether the existing stubs will do or not

sygnal/apnspushkin.py
H-Shay: I wondered about this but `opentracing/tags.py` definitely stated it was an int so I went with that: https://github.com/opentracing/opentracing-python/blob/3e1d357a348269ef54d67f761302fab93dbfc0f7/opentracing/tags.py#L108

It seems this is incorrect, I will change the type to match the implementation and not the comment.

DMRobertson: I'm guessing (haven't looked) that
- a span can be associated with multiple (tag, value) pairs.
- tag is a string identifying the meaning of the value. in this case, the string `'http.status_code'`
- value is probably an int!

H-Shay: Ahhh that makes a lot of sense!

H-Shay:
> 
> If you remove the lines from `mypy.ini` that skip opentracing and do `pip install types-opentracing`, does it lead to great success?
> 
It does indeed lead to great success. I will put up a separate PR with these changes, I suppose the best thing to do with this PR is to close/scrap it?

clokep:
It could be great to use those on Synapse too! Might be worth filing an issue.

sygnal/apnspushkin.py
reivilibre: I can see how the comment was a little misleading, but David's interpretation is what I believe also :)

reivilibre:
> It does indeed lead to great success. I will put up a separate PR with these changes

woot!

> I suppose the best thing to do with this PR is to close/scrap it?

Yes, probably. Sorry. But thanks for taking the time to do it anyway :)

DMRobertson:
> There appears to be a `types-opentracing` package (this is a common theme; worth checking for `types-xxx` and `xxx-stubs` when wanting stubs for a module `xxx`).

If it's any consolation, the package is apparently really new: first [released](https://pypi.org/project/types-opentracing/2.4.0/#history) last Thursday!

H-Shay:
> It could be great to use those on Synapse too! Might be worth filing an issue.

I can just do this today, not a problem.

clokep:
To cross reference, this was done in #287.

PR #284, opened 2021-12-06
Document that the `topic` is most commonly the Bundle Identifier for the iOS application.

reivilibre:
It came up in recent troubleshooting for a community member.
It would be useful to spell this out; from screenshots it appeared that the Apple developer site doesn't call it a topic, even though the APNs docs do.


squahtx (approved):
lgtm!

matrix.org's sygnal instance has a few `$bundle_id.voip` topics configured. I have no idea when they're used since my phone didn't register a token for one.

reivilibre:
> matrix.org's sygnal instance has a few $bundle_id.voip topics configured. I have no idea when they're used since my phone didn't register a token for one.

Yeah I don't really know what that is about. It might have been a historical detail from a previous version of APNs (they were there back when we switched to the new APNs with the Sygnal rewrite 2 years ago!), but I don't know with any certainty. Possibly a question for the iOS devs.

DMRobertson:
See https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns

> apns-topic
>  The  topic for the notification. In general, the topic is your app’s bundle  ID/app ID. It can have a suffix based on the type of push notification.  If you’re using a certificate that supports PushKit VoIP or watchOS  complication notifications, you must include this header with bundle ID  of you app and if applicable, the proper suffix. If you’re using  token-based authentication with APNs, you must include this header with  the correct bundle ID and suffix combination.

PR #285, opened 2021-12-06
Document the use of an iOS Notification Service Extension and the Push Gateway API as a workaround to trigger VoIP notifications on iOS.

reivilibre:
This came up as a recent community question, perfectly reasonably.

It's unfortunate that the workaround is required but as long as it is, it would be kind to document it.

### DIFF
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -175,6 +175,29 @@ An example `data` dictionary to specify on `POST /_matrix/client/r0/pushers/set`
 [APNs documentation]: https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html
 
 
+#### VoIP (voice calls and video calls) notifications on iOS
+
+iOS is capable of displaying an on-screen call notification (with answer/deny buttons)
+and playing a ringtone.
+
+However, this requires sending a special kind of push notification.
+Sygnal is not able to send this type of push notification because a VoIP call may
+begin as a result of an encrypted message, so Sygnal has no way to know that a given
+message should initiate a VoIP call.
+
+The current workaround is for your iOS application to register a Notification
+Service Extension and to use that Notification Service Extension to trigger a
+VoIP push notification with Sygnal when your application needs to display a call
+notification.
+This is done by calling [`/_matrix/push/v1/notify`](https://spec.matrix.org/latest/push-gateway-api/#post_matrixpushv1notify).
+(There is no interface to do this locally.)
+
+The Notification Service Extension of *Element*, [available here][ElementNSE] may be
+useful for reference.
+
+[ElementNSE]: https://github.com/vector-im/element-ios/blob/034e253fb19092ef16b5262293d5c32db96aec22/RiotNSE/NotificationService.swift
+
+
 ### Firebase Cloud Messaging
 
 The client will receive a message with an FCM `data` payload with this structure:
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -192,7 +192,7 @@ As of iOS 14.5, your iOS application can register a Notification Service Extensi
 upon receiving a relevant VoIP notification, calls [`reportNewIncomingVoIPPushPayload`][iOSReportVoIP]
 to trigger a VoIP notification display on the device.
 
-[iOSReportVoIP]
+[iOSReportVoIP]: https://developer.apple.com/documentation/callkit/cxprovider/3727263-reportnewincomingvoippushpayload
 
 
 ##### Old iOS (< 14.5) versions
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -185,11 +185,23 @@ Sygnal is not able to send this type of push notification because a VoIP call ma
 begin as a result of an encrypted message, so Sygnal has no way to know that a given
 message should initiate a VoIP call.
 
-The current workaround is for your iOS application to register a Notification Service Extension
+
+##### Recent iOS (≥ 14.5) versions
+
+As of iOS 14.5, your iOS application can register a Notification Service Extension that,
+upon receiving a relevant VoIP notification, calls [`reportNewIncomingVoIPPushPayload`][iOSReportVoIP]
+to trigger a VoIP notification display on the device.
+
+[iOSReportVoIP]
+
+
+##### Old iOS (< 14.5) versions
+
+For old iOS versions, a workaround is for your iOS application to register a Notification Service Extension
 that, upon receiving a relevant VoIP notification, makes an HTTP request to Sygnal's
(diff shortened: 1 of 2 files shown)


docs/applications.md
```diff
@@ -175,6 +175,29 @@ An example `data` dictionary to specify on `POST /_matrix/client/r0/pushers/set`
...
+Sygnal is not able to send this type of push notification because a VoIP call may
+begin as a result of an encrypted message, so Sygnal has no way to know that a given
+message should initiate a VoIP call.
+
+The current workaround is for your iOS application to register a Notification
+Service Extension and to use that Notification Service Extension to trigger a
+VoIP push notification with Sygnal when your application needs to display a call
+notification.
+This is done by calling [`/_matrix/push/v1/notify`](https://spec.matrix.org/latest/push-gateway-api/#post_matrixpushv1notify).
+(There is no interface to do this locally.)
```
squahtx, line 193: This last bit could be clearer. I wasn't expecting it to mean "your client manually asks sygnal to send a real notification back"

docs/applications.md
```diff
@@ -175,6 +175,29 @@ An example `data` dictionary to specify on `POST /_matrix/client/r0/pushers/set`
...
+
+However, this requires sending a special kind of push notification.
+Sygnal is not able to send this type of push notification because a VoIP call may
+begin as a result of an encrypted message, so Sygnal has no way to know that a given
+message should initiate a VoIP call.
+
+The current workaround is for your iOS application to register a Notification
+Service Extension and to use that Notification Service Extension to trigger a
+VoIP push notification with Sygnal when your application needs to display a call
+notification.
```
DMRobertson, line 191: I'm not even sure this is a workaround; IIRC this is intended on Apple's part.

I also vaguely remember the following from a previous role: If you send a voip call notification and don't present a CallKit UI within ~30seconds, then future voip call notifications will get delayed or throttled. I don't think this was a concretely documented thing other than "if you abuse this, we'll punish you".

I think links to Apple's documentation would be appropriate, e.g. https://developer.apple.com/documentation/pushkit/responding_to_voip_notifications_from_pushkit

DMRobertson: > If you send a voip call notification and don't present a CallKit UI within ~30seconds

I might have confused this for the section called Wake Your App with a Background Push [here](
https://developer.apple.com/documentation/backgroundtasks/choosing_background_strategies_for_your_app#overview).

reivilibre: > IIRC this is intended on Apple's part.

I would have thought the intended way would be to directly send the VoIP notifications to the app rather than have the app re-request one on demand

DMRobertson: Oh sorry, I didn't read properly. The use of the NSX to intercept and post-process/handle notification is definitely intentional; I didn't realise you were proposing that there'd be another request and another round-trip to the server.

docs/applications.md
```diff
@@ -175,6 +175,28 @@ An example `data` dictionary to specify on `POST /_matrix/client/r0/pushers/set`
...
+However, this requires sending a special kind of push notification.
+Sygnal is not able to send this type of push notification because a VoIP call may
+begin as a result of an encrypted message, so Sygnal has no way to know that a given
+message should initiate a VoIP call.
+
+The current workaround is for your iOS application to register a Notification Service Extension
+that, upon receiving a relevant VoIP notification, makes an HTTP request to Sygnal's
+[`/_matrix/push/v1/notify`](https://spec.matrix.org/latest/push-gateway-api/#post_matrixpushv1notify)
+endpoint in order to trigger the correct type of notification.
+(There is no interface to do this on-device.)
```
DMRobertson, line 192: I'm not 100% convinced of this. See e.g. https://developer.apple.com/documentation/callkit/sending_end-to-end_encrypted_voip_calls

reivilibre: That's an excellent find. Wonder if it's new? It doesn't appear on archive.org and the copyright year says 2021.

I will check with the Element iOS team if it would solve their woes in a less hacky way (also probably a more reliable way! :D).

DMRobertson: >  Wonder if it's new?

https://developer.apple.com/documentation/callkit/cxprovider/3727263-reportnewincomingvoippushpayload says it's available in iOS 14.5+, iPadOS 14.5+, Mac Catalyst 14.5+ . [Apparently](https://en.wikipedia.org/wiki/IOS_version_history#iOS_14_/_iPadOS_14) released end of April this year.

reivilibre: I opened an issue for Element iOS to make them aware: https://github.com/vector-im/element-ios/issues/5260

> developer.apple.com/documentation/callkit/cxprovider/3727263-reportnewincomingvoippushpayload says it's available in iOS 14.5+, iPadOS 14.5+, Mac Catalyst 14.5+ . Apparently released end of April this year.

Thanks for looking into it! I'll switch the documentation around to talk about this (maybe mentioning the workaround for historical iOS versions in the meantime for app developers that need to be compatible?).

reivilibre: No response thus far so I just documented it as best as I can.

docs/applications.md
```diff
@@ -175,6 +175,40 @@ An example `data` dictionary to specify on `POST /_matrix/client/r0/pushers/set`
...
+message should initiate a VoIP call.
+
+
+##### Recent iOS (≥ 14.5) versions
+
+As of iOS 14.5, your iOS application can register a Notification Service Extension that,
+upon receiving a relevant VoIP notification, calls [`reportNewIncomingVoIPPushPayload`][iOSReportVoIP]
+to trigger a VoIP notification display on the device.
+
+[iOSReportVoIP]
```
squahtx, line 195: This doesn't seem to render right in github?

reivilibre: urrrr.... I don't know why I didn't manage to actually write the link there. Thanks for spotting

docs/applications.md
```diff
@@ -175,6 +175,40 @@ An example `data` dictionary to specify on `POST /_matrix/client/r0/pushers/set`
 [APNs documentation]: https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html
 
 
+#### VoIP (voice calls and video calls) notifications on iOS
+
+iOS is capable of displaying an on-screen call notification (with answer/deny buttons)
+and playing a ringtone.
```
anoadragon453, line 181:
```suggestion
and play a ringtone.
```

reivilibre: 'iOS is capable of ... play a ringtone' doesn't sound right to me. How are you parsing this sentence in such a way that this is right? I can't easily spot it :)

anoadragon453: Oh! I have no idea. Ignore me!

DMRobertson (approved):

PR #287, opened 2021-12-06
Add `opentracing-types` to dev dependencies 

H-Shay:
As the title states, pulls in types for opentracing from `opentracing-types`. Has the added bonus of bumping the mypy precision rate to above 80%.

### DIFF
--- a/sygnal/sygnal.py
+++ b/sygnal/sygnal.py
@@ -153,7 +153,7 @@ class Sygnal:
                         scope_manager=AsyncioScopeManager(),
                     )
 
-                    self.tracer = jaeger_cfg.initialize_tracer()
+                    self.tracer = cast(Tracer, jaeger_cfg.initialize_tracer())
 
                     logger.info("Enabled OpenTracing support with Jaeger")
                 except ModuleNotFoundError:
--- a/sygnal/sygnal.py
+++ b/sygnal/sygnal.py
@@ -153,7 +153,9 @@ class Sygnal:
                         scope_manager=AsyncioScopeManager(),
                     )
 
-                    self.tracer = cast(Tracer, jaeger_cfg.initialize_tracer())
+                    jaeger_tracer = jaeger_cfg.initialize_tracer()
+                    assert jaeger_tracer is not None
+                    self.tracer = jaeger_tracer
 
                     logger.info("Enabled OpenTracing support with Jaeger")
                 except ModuleNotFoundError:
--- a/changelog.d/287.misc
+++ b/changelog.d/287.misc
@@ -1,0 +1,1 @@
+Add `opentracing-types` to the dev dependencies.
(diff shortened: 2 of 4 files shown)


DMRobertson (approved):
LGTM!

sygnal/sygnal.py
```diff
@@ -153,7 +153,7 @@ def __init__(
                         scope_manager=AsyncioScopeManager(),
                     )
 
-                    self.tracer = jaeger_cfg.initialize_tracer()
+                    self.tracer = cast(Tracer, jaeger_cfg.initialize_tracer())
```
DMRobertson, line 156: But having said that, any idea why this is necessary?

H-Shay: Honestly not entirely! I know in a previous conversation, there was discussion about `jaeger.Tracer` inheriting from `Tracer`: https://github.com/matrix-org/sygnal/pull/275#discussion_r752545187 but mypy doesn't seem to be picking that up and in a fit of pique I figured this was a pretty safe cast to make while fully expecting someone to call me on it.

reivilibre:
```
sygnal/sygnal.py:156: error: Incompatible types in assignment (expression has type "Optional[jaeger_client.tracer.Tracer]", variable has type "opentracing.tracer.Tracer")  [assignment]
```

I think the problem is that it's Optional.

Clicking through to the function, I see:
```python
    def initialize_tracer(self, io_loop: Optional[Any] = None) -> Optional[Tracer]:
        """
        Initialize Jaeger Tracer based on the passed `jaeger_client.Config`.
        Save it to `opentracing.tracer` global variable.
        Only the first call to this method has any effect.
        """

        with Config._initialized_lock:
            if Config._initialized:
                logger.warning('Jaeger tracer already initialized, skipping')
                return None
            Config._initialized = True

        tracer = self.new_tracer(io_loop)

        self._initialize_global_tracer(tracer=tracer)
        return tracer
```

I think my preferred approach would either to assert that it's not None,
or to only assign it to `self.tracer` if it's not None — logging an error and ignoring it if it *is* None (opentracing is not really critical to running Sygnal, so part of me feels that it'd be OK to start up without it if needed).

H-Shay: I thought that at first but when I added this code:
```
tracer = jaeger_cfg.initialize_tracer()
assert tracer is not None
self.tracer = tracer
```
the error persisted, which is why I thought it might be an issue with the inheritance not being clear to mypy.

reivilibre: I think that's because `tracer` already exists as a var with a different type annotation.

```python
                    tr = jaeger_cfg.initialize_tracer()
                    assert tr is not None
                    self.tracer = tr
```
seems to work locally, though probably give it a nicer name :D

H-Shay: Yes indeed, I have fixed it.

PR #288, opened 2021-12-08
Fix error when encountering web pushkeys with missing `endpoint`s

### LINKED ISSUE #257: Investigate Sentry error `Cannot use a string pattern on a bytes-like object`
First seen a month ago --- did we do some kind of release then?

Internal users: https://sentry.matrix.org/sentry/sygnal-matrixorg/issues/228100/?query=is%3Aunresolved

```
TypeError: cannot use a string pattern on a bytes-like object
  File "sygnal/http.py", line 265, in _handle_dispatch
    result = await pushkin.dispatch_notification(notif, d, context)
  File "sygnal/notifications.py", line 180, in dispatch_notification
    return await self._dispatch_notification_unlimited(n, device, context)
  File "sygnal/webpushpushkin.py", line 151, in _dispatch_notification_unlimited
    regex.fullmatch(endpoint_domain) for regex in self.allowed_endpoints
  File "sygnal/webpushpushkin.py", line 151, in <genexpr>
    regex.fullmatch(endpoint_domain) for regex in self.allowed_endpoints
```

Need to understand:

- which `regex` are we using and where did it come from
- ditto for endpoint_domain

squahtx:
Fixes #257

### DIFF
--- a/changelog.d/288.bugfix
+++ b/changelog.d/288.bugfix
@@ -1,0 +1,1 @@
+Reject web pushkeys with missing endpoints instead of raising an error.
--- a/changelog.d/288.bugfix
+++ b/changelog.d/288.bugfix
@@ -1,1 +1,1 @@
-Reject web pushkeys with missing endpoints instead of raising an error.
+Fix a bug introduced in Sygnal 0.9.1 where web pushkeys with missing endpoints would cause an error.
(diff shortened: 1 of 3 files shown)


clokep (approved):
This seems reasonable. Is `endpoint` defined anywhere in the spec (or our documentation)? Is it marked as required?

squahtx:
> Is `endpoint` defined anywhere in the spec (or our documentation)? Is it marked as required?

`devices.data` is pusher-specific according to the [spec](https://spec.matrix.org/v1.1/push-gateway-api/#post_matrixpushv1notify).
Our documentation notes that `endpoint` and `auth` are expected: https://github.com/matrix-org/sygnal/blob/main/docs/applications.md#push-key-and-expected-push-data

...and I've just noticed that there is already a check for `endpoint` and `auth` down below. The PR needs reworking.

clokep (approved):
Ah, that's a better fix! Sorry for missing the extra checks!

PR #290, opened 2021-12-14
Declare missing dependencies

squahtx:


squahtx:
The minimum package versions for attrs and typing-extensions are copied from Synapse: 
https://github.com/matrix-org/synapse/blob/eb39da6782b57c939450839097f32a14cba3ebfc/synapse/python_dependencies.py#L81-L85

The minimum package version for pyopenssl is copied from aioapns:
https://github.com/Fatal1ty/aioapns/blob/master/setup.py#L30

DMRobertson (commented):
Thanks for this!

DMRobertson:
Another thought: why did CI miss this? I guess there's no equivalent of the `olddeps` job that Synapse has?

setup.py
```diff
@@ -22,33 +22,33 @@
...
     "dev": [
-        "coverage~=5.5",
         "black==21.6b0",
+        "coverage~=5.5",
         "flake8==3.9.0",
         "isort~=5.0",
-        "mypy==0.812",
         "mypy-zope==0.3.0",
-        "tox",
+        "mypy==0.812",
```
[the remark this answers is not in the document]
squahtx: Good spot, I'll fix it.

callahad:
> Another thought: why did CI miss this? I guess there's no equivalent of the `olddeps` job that Synapse has?

I think we're not testing against older _python_ versions (e.g., I think setting up a matrix of `['3.7', '3.x']` would've found it)

setup.py
```diff
@@ -22,33 +22,36 @@
...
     "opentracing>=2.2.0",
+    "prometheus_client>=0.7.0,<0.8",
+    "py-vapid>=1.7.0",
+    "pyOpenSSL>=17.5.0",
+    "pywebpush>=1.13.0",
+    "pyyaml>=5.1.1",
     "sentry-sdk>=0.10.2",
+    "service_identity>=18.1.0",
+    "Twisted>=19.7",
+    "typing-extensions>=3.7.4",
```
callahad, line 41: Should we make this conditional so it's only installed on Py < 3.8? (And do the same for the import that needs it)

DMRobertson: I think we typically import directly from `typing_extensions`; would have to change the source to do a `try...except ImportError...` dance

reivilibre:
> Another thought: why did CI miss this? I guess there's no equivalent of the `olddeps` job that Synapse has?

Possibly because it was brought in by a dev dependency (e.g. `mypy`) and was present during the tests. :/

reivilibre:
> > Another thought: why did CI miss this? I guess there's no equivalent of the `olddeps` job that Synapse has?
> 
> I think we're not testing against older _python_ versions (e.g., I think setting up a matrix of `['3.7', '3.x']` would've found it)

3.7 is the oldest version we support iirc.

setup.py
```diff
@@ -23,6 +23,7 @@
 
 INSTALL_REQUIRES = [
     "aioapns>=1.10",
+    "attrs>=19.2.0",
     "cryptography>=2.6.1",
     "idna>=2.8",
     "importlib_metadata",
```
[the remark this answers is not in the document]
squahtx: Perhaps. I have no idea what the lower bound should be

setup.py
squahtx: On the import side, mypy does not appreciate the obvious attempt:
https://mypy-play.net/?mypy=latest&python=3.10&gist=9ac99f23480e5c8bc86c68f94ce65b6c

```py
if TYPE_CHECKING or sys.version_info < (3, 8, 0):
    from typing_extensions import Literal
else:
    from typing import Literal
```
works with mypy but always requires typing_extensions in dev.

setup.py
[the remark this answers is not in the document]
squahtx: For reference, this is what we do in python-signedjson: https://github.com/matrix-org/python-signedjson/pull/9

setup.py
squahtx: There's a PR to do the same for python-signedjson here: https://github.com/matrix-org/python-signedjson/pull/12

DMRobertson (approved):
I'm happy if CI is happy. Sanity check: would this have fixed the earlier deployment to matrix.org?

squahtx:
> I'm happy if CI is happy. Sanity check: would this have fixed the earlier deployment to matrix.org?

It ought to, assuming we do a `pip install` as part of the deployment.
I have no idea how we'd test this apart from doing another deployment.

DMRobertson:
> I have no idea how we'd test this apart from doing another deployment.

Maybe a `pip install -e .`? But yeah, probably best to try it and see.

squahtx:
> Maybe a `pip install -e .`? But yeah, probably best to try it and see.

A `pip install -r <(./scripts-dev/old_deps.py)` seems to be okay.

--------------------------------

On another note, mypy really does not like conditional imports. Getting the right combination of `type: ignore`s is far from simple.

Python 3.7:
```
sygnal/__init__.py:17: error: Module 'importlib.metadata' has no attribute 'PackageNotFoundError'  [attr-defined]
sygnal/__init__.py:17: error: Module 'importlib.metadata' has no attribute 'version'  [attr-defined]
```

Python 3.8, importlib_metadata installed:
```
sygnal/__init__.py:22: error: Incompatible import of "PackageNotFoundError" (imported name has type "Type[importlib_metadata.PackageNotFoundError]", local name has type "Type[importlib.metadata.PackageNotFoundError]")  [misc]
```

Python 3.8, importlib_metadata not installed:
```
sygnal/__init__.py:22: error: Name 'PackageNotFoundError' already defined (possibly by an import)  [no-redef]
sygnal/__init__.py:22: error: Name 'version' already defined (possibly by an import)  [no-redef]
```

DMRobertson:
> On another note, mypy really does not like conditional imports. Getting the right combination of `type: ignore`s is far from simple.

Ouch, thanks for working through that. Is there any way that we could use `TYPE_CHECKING` to help? I guess not, since it's sensitive to the version of the stdlib that's running mypy(?)

squahtx:
> Ouch, thanks for working through that. Is there any way that we could use `TYPE_CHECKING` to help? I guess not, since it's sensitive to the version of the stdlib that's running mypy(?)

I can't see how, unless we unconditionally use importlib_metadata when type checking.


and you've made me realise that the dependencies are wrong. typing_extensions is only installed on Python <3.8, but we use it in type checking mode regardless.

squahtx:
mypy brings in typing-extensions for us, but we should be explicit.

DMRobertson:
> > I'm happy if CI is happy. Sanity check: would this have fixed the earlier deployment to matrix.org?
> 
> It ought to, assuming we do a `pip install` as part of the deployment. I have no idea how we'd test this apart from doing another deployment.

Ahhh, I was more getting at "why did the deployment fail?" so we could check this fixes it. The cause was an inability to import typing_extensions, but that's now been made a hard requirement. 

So I think this is hopefully good to go! :+1:

PR #292, opened 2022-01-05
Check if `default_payload` is a dict before using it as a dict

### LINKED ISSUE #291: `TypeError: 'NoneType' object is not iterable` in apnspushkin.py and gcmpushkin.py
In gcmpushkin.py and apnspushkin.py, we attempt to use `device.data["default_payload"]` as a dictionary without verifying its type. Since `device.data` is arbitrary client or homeserver-supplied JSON, this can lead to errors. We ought to check the type of  `device.data["default_payload"]` before using it:
https://github.com/matrix-org/sygnal/blob/v0.11.0/sygnal/gcmpushkin.py#L425-L426
https://github.com/matrix-org/sygnal/blob/v0.11.0/sygnal/apnspushkin.py#L353-L354

For reference, webpushpushkin.py does include a check:
https://github.com/matrix-org/sygnal/blob/v0.11.0/sygnal/webpushpushkin.py#L245-L248

H-Shay:
Fixes #291.


sygnal/apnspushkin.py
```diff
@@ -351,7 +351,9 @@ def _get_payload_event_id_only(
         payload = {}
 
         if device.data:
-            payload.update(device.data.get("default_payload", {}))
+            default_payload = device.data.get("default_payload")
+            if isinstance(default_payload, dict):
+                payload.update(default_payload)
```
DMRobertson, line 356: This change makes us ignore any `default_payload`s which aren't a dict. Is providing such a `default_payload` legitimate, or should we consider it a request error (e.g. HTTP 400 Bad Request)?

The same for gcm pushkin below.

changelog.d/292.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in Sygnal 0.7.0 where a dict method is called on an object before it is determined to be a dict, causing TypeErrors.
```
DMRobertson, line 1: I think we usually try to write these from the perspective of the user. Can we describe what triggered this and the previous broken behaviour (a 500 internal server error?)

sygnal/apnspushkin.py
clokep: Note that the webpush pushking already does this exact code.

H-Shay: The Sygnal docs do specify that `default_payload` should be a dict, (although I can't find anything about `default_payload` in the spec fwiw): https://github.com/matrix-org/sygnal#pusher-data-configuration

I think it's fair to consider anything other than a dict a Bad Request and throw an error in those cases, if people agree with this I can also change the behavior of the webpush pushkin to reflect that as well.

squahtx: If we decide to go down that route, I'd rather we reject the pushkey (tell the homeserver to forget that pushkey+device data combination) rather than 400. Otherwise the homeserver will continue to send us notifications that we refuse to handle.

H-Shay: It sounds like some of the options to handle `default_payloads` of the incorrect type are:
-throw a bad request error and 400 
-reject the pushkey
-drop the `default_payload` data silently (what's currently happening in the PR)

I'm trying to think about what would make the most sense/be least annoying to handle for the end user. Would it make sense to ignore the `default_payload` if it is malformed but warn about that in the logs?

clokep: An invalid `default_payload` is a programming error, I believe, so rejecting the `pushkey` probably makes the most sense?

The push spec is a bit confusing since much of it is left up to implementation details between client app and push server (with the homeserver just proxying between).

H-Shay: Alright I will reject the `pushkey`, thanks for everyone's input!

DMRobertson:
I think this is waiting on changes to reject the pushkey(?)

H-Shay:
Yes I've add changes which reject the pushkey-I am not convinced of their elegance but I figured best to get some feedback on my approach. Once the `gcmpushkin` and `apsnpushkin` changes look good, I can change `webpushkin` as well.

sygnal/apnspushkin.py
```diff
@@ -283,6 +283,15 @@ async def _dispatch_notification_unlimited(
         with self.sygnal.tracer.start_span(
             "apns_dispatch", tags=span_tags, child_of=context.opentracing_span
         ) as span_parent:
+            # Before we build the payload, check that the default_payload is not
+            # malformed and reject the pushkey if it is
+            if device.data:
+                default_payload = device.data.get("default_payload", {})
+                if not isinstance(default_payload, dict):
+                    log.error(
+                        "Default_payload is malformed, this value must be a dict."
```
DMRobertson, line 292: I'd recommend not capitalising the `d` in `default` here, since `default_payload` is a (case sensitive) key.

sygnal/apnspushkin.py
```diff
@@ -351,7 +360,8 @@ def _get_payload_event_id_only(
         payload = {}
 
         if device.data:
-            payload.update(device.data.get("default_payload", {}))
+            default_payload = device.data.get("default_payload", {})
+            payload.update(default_payload)
```
DMRobertson, line 364: Does this use of `default_payload` need checking for not being a dict too? (Not sure of the bigger picture here)

sygnal/gcmpushkin.py
```diff
@@ -347,16 +347,24 @@ async def _dispatch_notification_unlimited(
         with self.sygnal.tracer.start_span(
             "gcm_dispatch", tags=span_tags, child_of=context.opentracing_span
         ) as span_parent:
+            # TODO: Implement collapse_key to queue only one message per room.
+            failed: List[str] = []
+
             data = GcmPushkin._build_data(n, device)
+
+            # Reject pushkey if default_payload is misconfigured
+            if not data:
```
DMRobertson, line 356: We will hit this branch if `data is None` (good), but also if `data` is an empty dict (maybe bad?) because empty dictionaries are falsey:

```python
>>> bool({})
False
```

Is this intended? If so, the error message is confusing.

(Prefer `x is None` or `x is not None` when testing for none-ness)

sygnal/gcmpushkin.py
```diff
@@ -347,16 +347,24 @@ async def _dispatch_notification_unlimited(
...
+                log.error(
+                    "Default_payload was misconfigured, this value must be a dict."
+                )
```
DMRobertson, line 359: I'd be tempted to log in `_build_data` if possible, so that the log message is next to the logic that detects the error condition.

DMRobertson (commented):
A few things.

Thank you for the test cases! Is it worth having one for the webpushkin too?

sygnal/apnspushkin.py
H-Shay: ah that makes sense and I never would have thought of that, thanks for pointing out.

sygnal/apnspushkin.py
H-Shay: My thinking on this was that we do the check on `default_payload` several lines before the only call to the function where this code is located (`_get_payload_event_id_only`), and the check returns/stops execution if the default payload is not a dict, so afaict the execution should never get to this stage if `default_payload` is not a dict.

sygnal/gcmpushkin.py
H-Shay: I should have tested for None, have fixed.

sygnal/apnspushkin.py
DMRobertson: I think your reasoning is correct; but I have in mind:

- maintainability: in the future, we might call `_get_payload_event_id_only` with a `device` whose `default_payload` hasn't been validated
- ease of reading: it'd be nice to arrange this in a way that the reader doesn't need to repeat this reasoning every time we come to look at `default_payload`.

(Warning: the following is splitting hairs.)

Looking at the body of `_get_payload_event_id_only`, I note that the only time we use the `device` argument is to retrieve the `default_payload` (see also https://github.com/matrix-org/sygnal/pull/127). I'm tempted to suggest:

- remove `_get_payload_event_id_only`'s `device` argument
- have it take a `default_payload: Dict[str, Any]` instead
- change the caller to pass `default_payload` directly, handling the case where `device.data is None`

H-Shay: Great I will do that, thanks for outlining your reasoning!

DMRobertson: (To emphasise: this is fine as it is and I _am_ going a little too extreme here. I think this is great as written and definitely mergeable. But I wanted to outline the thinking behind the question.)

DMRobertson: Ahh, I had something slightly different in mind. I was thinking:

- above line 288 add `default_payload = {}`
- this means we can remove lines 297 through 301, and always use 302 (because we _always_ have a default payload)

H-Shay: Right that makes more sense, have updated.

DMRobertson (approved):
LGTM, and apologies again for my nitpicking.

sygnal/gcmpushkin.py
```diff
@@ -347,16 +347,21 @@ async def _dispatch_notification_unlimited(
...
             "gcm_dispatch", tags=span_tags, child_of=context.opentracing_span
         ) as span_parent:
+            # TODO: Implement collapse_key to queue only one message per room.
+            failed: List[str] = []
+
             data = GcmPushkin._build_data(n, device)
+
+            # Reject pushkey if default_payload is misconfigured
+            if data is None:
+                failed.append(device.pushkey)
```
richvdh, line 357: looks like this will go on to attempt delivery anyway?

H-Shay: It sure does look like that-which probably doesn't make sense. I can fix that along with downgrading the error to a warning.

PR #294, opened 2022-01-27
Don't use aioapns 2.1

### LINKED ISSUE #293: build failed due to aioapns dependency version
https://github.com/matrix-org/sygnal/blob/bd1c408902988c624ee6c87e97ec6af454698156/setup.py#L25

aioapns 2.1 introduced a breaking change. (remove loop on APNS instantiation).

to fix it, set the dep to : 

    "aioapns==2.0.2",

DMRobertson:
Fixes #293.

aioapns 2.1 made a breaking change: one can no longer pass an event loop directly to APNs.__init__. We could change our code to fix it, but (regrettably) I think it's quicker to avoid using that version. See also https://github.com/matrix-org/sygnal/issues/293#issuecomment-1023453460

I wonder if unit tests would have spotted this? Presumably not.

### DIFF
--- a/changelog.d/294.misc
+++ b/changelog.d/294.misc
@@ -1,0 +1,1 @@
+Avoid a breaking change in aioapns 2.1 by requiring an earlier version of that package.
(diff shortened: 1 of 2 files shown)


callahad (dismissed):

DMRobertson:
Oh bugger, newsfile.

PR #305, opened 2022-06-09
Mark iOS push notifications with `push_type` = `alert` to allow for on device silencing

stefanceriu:
We were recently granted the [com.apple.developer.usernotifications.filtering](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_usernotifications_filtering) entitlement which allows push notifications to be completely discarded once received on the device. This is useful for hiding notifications from encrypted events we don't care about like poll answers (as opposed to showing an empty notification).

One caveat is that it only works for `alert` `apns-push-type` notifications which is an optional field we're not currently setting.
 
This PR does just that through the use of aioapns's [PushType](https://github.com/Fatal1ty/aioapns/blob/master/aioapns/common.py#L9) [header](https://github.com/Fatal1ty/aioapns/blob/master/aioapns/connection.py#L187).

### DIFF
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -24,7 +24,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 from uuid import uuid4
 
 import aioapns
-from aioapns import APNs, NotificationRequest
+from aioapns import APNs, NotificationRequest, PushType
 from aioapns.common import NotificationResult
 from cryptography.hazmat.backends import default_backend
 from cryptography.x509 import load_pem_x509_certificate
@@ -230,6 +230,7 @@ class ApnsPushkin(ConcurrencyLimitedPushkin):
             message=shaved_payload,
             priority=prio,
             notification_id=notif_id,
+            push_type=PushType.ALERT,
         )
 
         try:
--- a/changelog.d/305.misc
+++ b/changelog.d/305.misc
@@ -1,0 +1,1 @@
+Mark iOS push notifications with `push_type` = `alert` to allow for on device silencing.


clokep:
What happens if an app isn't approved for this and we set this push type?

sygnal/apnspushkin.py
```diff
@@ -230,6 +230,7 @@ async def _dispatch_request(
             message=shaved_payload,
             priority=prio,
             notification_id=notif_id,
+            push_type=PushType.ALERT,
```
squahtx, line 233: The apple docs imply that if we use the alert push type, we have to use the non-voip topic:
https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns/

> **alert**
> Use the alert push type for notifications that trigger a user interaction—for example, an alert, badge, or sound. If you set this push type, the apns-topic header field must use your app’s bundle ID as the topic. For more information, see Generating a remote notification.
>
> **voip**
> [...] If you set this push type, the apns-topic header field must use your app’s bundle ID with .voip appended to the end.

However it seems that we configure sygnal to use both voip and non-voip topics:
```
    - name: im.vector.app.ios.prod
      [...]
      topic: im.vector.app

    - name: im.vector.app.ios.dev
      platform: sandbox
      [...]
      topic: im.vector.app

    - name: im.vector.app.ios.voip.prod
      [...]
      topic: im.vector.app.voip

    - name: im.vector.app.ios.voip.dev
      platform: sandbox
      [...]
      topic: im.vector.app.voip
```

I think it's best that we make a push_type config option that defaults to `None`.
We could then roll it out to dev builds first, before turning it on for production builds.

stefanceriu:
This PR is incomplete and would break voip UX. Raised separate ticket here https://github.com/matrix-org/sygnal/issues/308

Issue #308, opened 2022-06-10
Support for on device notification filtering


stefanceriu:
We were recently granted the [com.apple.developer.usernotifications.filtering](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_usernotifications_filtering) entitlement which allows push notifications to be completely discarded once received on the device. This is useful for hiding notifications from encrypted events we don't care about like poll answers (as opposed to showing an empty notification).

One caveat is that it only works for alert apns-push-type notifications which is an optional field we're not currently setting.

Based of https://github.com/matrix-org/sygnal/pull/305 we realized that setting that on all notifications would most likely break the voip ones as if the header doesn't match the topic ["APNs may return an error, delay the delivery of the notification, or drop it altogether."](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns/)

When a call invite "alert" notification comes in the iOS app makes a request to Sygnal to send a "voip" notification which in turn shows the full screen incoming call. This is done by sending a different appId in the request payload that matches one of the .voip configurations.

One reasonable way to go forward is to make the `apns-push-type` part of the Sygnal configuration and have it set differently per appId/topic.

PR #309, opened 2022-06-28
Add support for setting the apns-push-type header via device data

N-Pex:
Signed-off-by: N-Pex <npexdev@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,8 @@ For either type, it can accept:
 - the `platform` parameter which determines whether the production or sandbox
   APNS environment is used.
   Valid values are 'production' or 'sandbox'. If not provided, 'production' is used.
-
+- the `push_type` parameter which determines what 'apns-push-type' header is sent to
+  APNS. If not provided, the header is not sent.
 
 ### gcm
 
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ pusher (see
 
 See the sample configuration for examples.
 
+You can find a docker image for sygnal [on DockerHub](https://hub.docker.com/r/matrixdotorg/sygnal).
+
 
 App Types
 ---------
--- a/README.md
+++ b/README.md
@@ -72,8 +72,8 @@ For either type, it can accept:
 - the `platform` parameter which determines whether the production or sandbox
   APNS environment is used.
   Valid values are 'production' or 'sandbox'. If not provided, 'production' is used.
-- the `push_type` parameter which determines what 'apns-push-type' header is sent to
-  APNS. If not provided, the header is not sent.
+- the `push_type` parameter which determines what value for the `apns-push-type` header is sent to
+  APNs. If not provided, the header is not sent.
 
 ### gcm
 
--- a/changelog.d/309.feature
+++ b/changelog.d/309.feature
@@ -1,0 +1,1 @@
+Add support for setting the 'apns-push-type' header via device data. This is needed if notifications are to be silenced client side.
--- a/changelog.d/309.feature
+++ b/changelog.d/309.feature
@@ -1,1 +1,1 @@
-Add support for 'push_type' option in app config. This is needed if notifications are to be silenced client side.
+Add a new `push_type` configuration option for APNs apps, to control the value of the `apns-push-type` header when sending requests.
--- a/changelog.d/309.feature
+++ b/changelog.d/309.feature
@@ -1,1 +1,1 @@
-Add support for setting the 'apns-push-type' header via device data. This is needed if notifications are to be silenced client side.
+Add support for 'push_type' option in app config. This is needed if notifications are to be silenced client side.
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -73,10 +73,12 @@ CERTIFICATE_EXPIRATION_GAUGE = Gauge(
     labelnames=["pushkin"],
 )
 
+
 class ApnsPushTypeHeader:
     def __init__(self, value):
         self.value = value
 
+   
 class ApnsPushkin(ConcurrencyLimitedPushkin):
     """
     Relays notifications to the Apple Push Notification Service.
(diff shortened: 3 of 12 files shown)


squahtx:
There was a previous attempt at this feature in #305, where we considered making the push type a per-pusher config option.

N-Pex:
Oh, interesting. We came across this requirement quite a while back. I haven't come around to implementing it until now.

babolivier:
Related: https://github.com/matrix-org/sygnal/issues/308

babolivier:
Github why did you close this PR, I did not click that button.

babolivier:
I don't think this is how we want to fix this issue. The device data is defined in the request to create the pusher, which format is defined in the spec: https://spec.matrix.org/latest/client-server-api/#post_matrixclientv3pushersset

I'm uncomfortable relying on a key that isn't described in the specification, in a structure that is. So if we were to implement it with device data we would need a companion MSC (Matrix Spec Change, a proposal to update the Matrix specification) to go with it. And since it's very specific to one platform I'm not sure how likely it would be to be accepted.

On the other hand, https://github.com/matrix-org/sygnal/issues/308 describes an alternative solution using the app configuration, which gets you basically the same result with a bit less trouble, and also has the added bonus of making the fix work on already existing pushers (whereas if we were using device data, existing clients would need to re-register their pushers). This is definitely a solution we'd be happier with.

@N-Pex would you be happy to change your PR so that it instead implements the solution described in https://github.com/matrix-org/sygnal/issues/308? (if not I'll have a look next week)

babolivier:
Gosh darn it Github.

N-Pex:
Thanks @babolivier ! Yeah, I think that makes a lot of sense! I'll have a stab at it, but if I fail (given my almost total lack of Python skills, haha) I might need you to chime in. I'll get back to you soon here =)

N-Pex:
Ok, changes pushed! Not sure what to call the option, but I thought "push_type" would be proper?

N-Pex:
Sorry about that, I guess I should lint this all, but my tooling is not properly setup.

sygnal/apnspushkin.py
```diff
@@ -188,6 +194,12 @@ def __init__(self, name: str, sygnal: "Sygnal", config: Dict[str, Any]) -> None:
                 loop=loop,
             )
 
+        push_type = self.get_config("push_type", str)
+        if not push_type:
+            self.push_type = None
+        else:
+            self.push_type = ApnsPushTypeHeader(push_type)
```
babolivier, line 201: As far as I understand it, aioapns expects an `aioapns.PushType` (which is an `Enum`) here. So I think something we should do is: first check whether the value of `push_type` is a member of this enum, and then use the matching value within the enum for `self.push_type`.

I think this is best done by declaring a constant which is a dictionary between a string value and its matching enum member (because annoyingly there isn't a way to list all members of an enum in Python without doing some voodoo with its internals). So something like this:

```python
APNS_PUSH_TYPES = {
    "alert": PushType.ALERT,
    "background": PushType.BACKGROUND,
    ...
}
```

(you can see the list of members of `PushType` here: https://github.com/Fatal1ty/aioapns/blob/26c2cdc3aacb529f4de1065bf3d1cff2137f13bc/aioapns/common.py#L9-L15 - unfortunately aioapns doesn't seem to have docs so we need to refer to the code)

Then you can do something like:

```python
push_type = self.get_config("push_type", str)
if not push_type:
    self.push_type = None
else:
    if push_type not in APNS_PUSH_TYPES.keys():
        raise PushkinSetupException(f"Invalid value for push_type: {push_type}")

    self.push_type = APNS_PUSH_TYPES[push_type]
```

Which a) ensures we use a supported value, b) removes the need for an internal class in favour of one provided by aioapns.

README.md
```diff
@@ -72,7 +72,8 @@ For either type, it can accept:
 - the `platform` parameter which determines whether the production or sandbox
   APNS environment is used.
   Valid values are 'production' or 'sandbox'. If not provided, 'production' is used.
-
+- the `push_type` parameter which determines what 'apns-push-type' header is sent to
```
babolivier, line 75:
```suggestion
- the `push_type` parameter which determines what value for the `apns-push-type` header is sent to
```

README.md
```diff
@@ -72,7 +72,8 @@ For either type, it can accept:
...
+  APNS. If not provided, the header is not sent.
```
babolivier, line 76:
```suggestion
  APNs. If not provided, the header is not sent.
```

Annoyingly Apple has decided to be very irritating with the official casing of `APNs`.

changelog.d/309.feature
```diff
@@ -0,0 +1 @@
+Add support for 'push_type' option in app config. This is needed if notifications are to be silenced client side.
```
babolivier, line 1:
```suggestion
Add a new `push_type` configuration option for APNs apps, to control the value of the `apns-push-type` header when sending requests.
```

babolivier: Also feel free to credit yourself by adding a `Contributed by ...` (with your name, username or anything else) at the end if you wish to do so :)

babolivier (changes requested):
Thanks for your work! I've noted a few things that I'd like to see improved. I'd also like to see a test for this (you can use the tests [here](https://github.com/matrix-org/sygnal/blob/80f1d4820f248730cff626974496c0960b0d010c/tests/test_apns.py) for inspiration).

Feel free to drop by [`#sygnal:matrix.org`](https://matrix.to/#/#sygnal:matrix.org) if you need help with this too! :)

N-Pex:
Ok, I think I got most of it done (well, you wrote almost all the code, haha) =)

babolivier (approved):
Fantastic, thanks a lot for your contribution! :)

N-Pex:
Thanks for the quick responses and the guidance, glad we could get this in!

PR #312, opened 2022-07-05
Add TTL support for FCM notifications

Fizzadar:


squahtx:
`Notification` fields seem to be defined in the spec. We might need a proposal to add a `ttl` field?

squahtx (commented):
The `Notification` dict is specced and it's impolite for us to add extra fields to it.

I think we're allowed to put pusher-specific fields in `device.data`, so that could be an option?

PR #315, opened 2022-07-26
Don't attempt delivery of notification if we have rejected pushkey

### LINKED ISSUE #314: "default_payload was misconfigured, this value must be a dict" error is noisy
**Describe the bug**
As of https://github.com/matrix-org/sygnal/pull/292 we now check that `default_payload` is a dictionary before interacting with it. `default_payload` is a user-supplied dictionary which is set when a client sets up their pusher via [`POST /_matrix/client/v3/pushers/set`](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3pushersset).

While we now successfully catch when clients set this value incorrectly, raising an error means that this we get notified via Sentry when it happens - and as this data is user-supplied, we can't really do much about it, making the error noise.

This issue proposes to downgrade the log from an error to a warning. Such instances will then still be logged, but not show up in Sentry/potentially alarm sysadmins.

H-Shay:
Previously we decided to reject pushkeys if `default_payload` was malformed (i.e. not a `dict`) but went ahead and attempted delivery with the rejected pushkey anyway. This PR fixes that and changes the code so we do not attempt delivery with a pushkey we have rejected. 

While I was in the area I also downgraded the logging error messages to a warning, to reduce sentry spam. This fixes #314. I also amended the warnings to give a little more information on why the pushkey was being rejected.


sygnal/apnspushkin.py
```diff
@@ -311,8 +311,9 @@ async def _dispatch_notification_unlimited(
             if device.data:
                 default_payload = device.data.get("default_payload", {})
                 if not isinstance(default_payload, dict):
-                    log.error(
-                        "default_payload is malformed, this value must be a dict."
+                    logger.warning(
+                        "Rejecting pushkey due to misconfigured default_payload, "
+                        "please ensure that default_payload is a dict."
                     )
```
squahtx, line 317: I'm confused about the difference between `log` and `logger`. What's the difference between them?

tests/test_gcm.py
```diff
@@ -142,7 +142,6 @@ def test_misformed_default_payload_rejected(self):
         )
 
         self.assertEqual(resp, {"rejected": ["badpayload"]})
-        self.assertEqual(gcm.num_requests, 1)
```
squahtx, line 145: Should we keep this and assert that there were 0 requests instead?

sygnal/gcmpushkin.py
```diff
@@ -354,7 +354,14 @@ async def _dispatch_notification_unlimited(
...
             # Reject pushkey if default_payload is misconfigured
             if data is None:
+                logger.warning(
+                    "Rejecting pushkey due to misconfigured default_payload, "
+                    "please ensure that default_payload is a dict."
+                )
                 failed.append(device.pushkey)
+                pushkeys.remove(device.pushkey)
+                if len(pushkeys) == 0:
+                    return failed
```
squahtx, line 364: We're about to use the same `data` for all the pushkeys in `pushkeys`, so I think we have to reject all of them instead of just one?

sygnal/apnspushkin.py
H-Shay: Looks like I was confused too, but after looking at this, `logger` is the original logger which is instantiated on line 70, and gets passed in to `_dispatch_notification_unlimited` where it is used to instantiate `log`, which is a `NotificationLoggerAdaptor`, on line 325. So line 314 should really still use `log` instead of `logger`, thanks for catching that.

sygnal/gcmpushkin.py
H-Shay: Right that makes sense.

sygnal/apnspushkin.py
```diff
@@ -311,8 +311,9 @@ async def _dispatch_notification_unlimited(
             if device.data:
                 default_payload = device.data.get("default_payload", {})
                 if not isinstance(default_payload, dict):
-                    log.error(
-                        "default_payload is malformed, this value must be a dict."
+                    logger.warning(
```
squahtx, line 314: I think we want to continue using `log` instead of `logger` here too?

H-Shay: Yes I missed this, have fixed.

H-Shay:
Merged, thanks for the review!

PR #319, opened 2022-08-03
add data["devices"] for Fluffychat messenger

taherfattahi:
I tested FluffyChat application and I realized we need to add this line of code for working with FCM push notification

### DIFF
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -467,7 +467,7 @@ class GcmPushkin(ConcurrencyLimitedPushkin):
             data["unread"] = n.counts.unread
             data["missed_calls"] = n.counts.missed_calls
         
-        # this line tested for Fluffychat application 
+        # this line tested for Fluffychat 
         data["devices"] = [{"pushkey": device.pushkey,"data": device.data,"pushkey_ts": device.pushkey_ts,"app_id": device.app_id}]
 
         return data
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -466,5 +466,8 @@ class GcmPushkin(ConcurrencyLimitedPushkin):
         if getattr(n, "counts", None):
             data["unread"] = n.counts.unread
             data["missed_calls"] = n.counts.missed_calls
+        
+        # this line tested for Fluffychat application 
+        data["devices"] = [{"pushkey": device.pushkey,"data": device.data,"pushkey_ts": device.pushkey_ts,"app_id": device.app_id}]
 
         return data
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -468,6 +468,6 @@ class GcmPushkin(ConcurrencyLimitedPushkin):
             data["missed_calls"] = n.counts.missed_calls
         
         # this line tested for Fluffychat 
-        data["devices"] = [{"pushkey": device.pushkey,"data": device.data,"pushkey_ts": device.pushkey_ts,"app_id": device.app_id}]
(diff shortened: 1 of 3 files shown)


sygnal/gcmpushkin.py
```diff
@@ -466,5 +466,8 @@ def _build_data(n: Notification, device: Device) -> Optional[Dict[str, Any]]:
         if getattr(n, "counts", None):
             data["unread"] = n.counts.unread
             data["missed_calls"] = n.counts.missed_calls
+        
+        # this line tested for Fluffychat 
+        data["devices"] = [{"pushkey": device.pushkey,"data": device.data,"pushkey_ts": device.pushkey_ts,"app_id": device.app_id}]
```
richvdh, line 471: I don't have particular objections to adding extra fields to support fluffychat, but:

1. I think it should be configurable (in `sygnal.yaml`). There's no point adding extra data on pushes to apps that aren't going to use it.
2. I'd rather like to understand the reasons fluffychat needs this data. Shouldn't the app already know this information? (Also, why is it an array `devices` instead of a single `device`)?
3. Any changes made here will need documenting in https://github.com/matrix-org/sygnal/blob/main/docs/applications.md

taherfattahi: FluffyChat FCM push notification can work with that line of code,
ok, thanks a lot for your response 🙏

richvdh:
@taherfattahi I don't think we can accept this in is current state. If you're interested in making the changes I requested, let us know, but I'm going to close this for now.

vanyasem:
This isn't directly related to FluffyChat, but to the underlying `matrix` library implementation in Dart: https://pub.dev/packages/matrix

The `PushNotification` object in the library has a mandatory `devices` field: https://github.com/famedly/matrix-dart-sdk/blob/v0.34.0/lib/src/utils/push_notification.dart#L20

I believe they decided to make it required because of the following text in the matrix push gateway API spec: https://spec.matrix.org/v1.2/push-gateway-api/


I believe it's not used in any capacity by the library

vanyasem:
Filed an issue in `matrix-dart-sdk` repo: https://github.com/famedly/matrix-dart-sdk/issues/1947

vanyasem:
`unread` and `missed_calls` fields also appear to be incompatible with `matrix-dart-sdk` (as they are wrapped in a `counts` object). Filed an issue: https://github.com/famedly/matrix-dart-sdk/issues/1948

vanyasem:
Here's an example of how to workaround both issues on the client-side without breaking compatibility with other Push Gateway implementations:

```dart
Future<void> _handlePushMessage(Map<dynamic, dynamic> messageData) {
  // Workaround: https://github.com/famedly/matrix-dart-sdk/issues/1947
  const String devicesKey = 'devices';
  if (!messageData.containsKey(devicesKey)) {
    messageData[devicesKey] = <PushNotificationDevice>[];
  }

  // Workaround: https://github.com/famedly/matrix-dart-sdk/issues/1948
  const String countsKey = 'counts';
  const String unreadKey = 'unread';
  const String missedCallsKey = 'missed_calls';
  if (!messageData.containsKey(countsKey) &&
      (messageData.containsKey(unreadKey) ||
          messageData.containsKey(missedCallsKey))) {
    messageData[countsKey] = PushNotificationCounts(
      unread: int.tryParse(messageData[unreadKey]),
      missedCalls: int.tryParse(messageData[missedCallsKey]),
    ).toJson();
  }

  final PushNotification notification = PushNotification.fromJson(
    Map<String, dynamic>.from(messageData),
  );
  // The rest of the logic
}
```

PR #331, opened 2023-03-16
Fix FCM second retry sends

Fizzadar:
When the first attempt succeeds for all but one pushkey the second attempt will end up populating both the `to` and `registration_ids` fields, which is rejected by FCM (400). For context, full errors looked like:

```
"[...] 400 from server, we have sent something invalid! Error: 'Must use either \"registration_ids\" field or \"to\", not both\\n'"
```

### DIFF
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -371,10 +371,13 @@ class GcmPushkin(ConcurrencyLimitedPushkin):
             body["priority"] = "normal" if n.prio == "low" else "high"
 
             for retry_number in range(0, MAX_TRIES):
+                # Should only ever be one of `to` or `registration_ids` fields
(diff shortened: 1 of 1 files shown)


sygnal/gcmpushkin.py
```diff
@@ -371,10 +371,13 @@ async def _dispatch_notification_unlimited(
             body["priority"] = "normal" if n.prio == "low" else "high"
 
             for retry_number in range(0, MAX_TRIES):
+                # Should only ever be one of `to` or `registration_ids` fields
```
clokep, line 374: Would it be clearer to rebuild the `body` at the top of this loop?

Any chance of being able to test this?

clokep (changes requested):
This needs lint & newsfragment also!

PR #336, opened 2023-04-23
Improve static type checking

dklimpel:
Signed-off-by: Dirk Klimpel [dirk@klimpel.org](mailto:dirk@klimpel.org)

### DIFF
--- a/sygnal/helper/context_factory.py
+++ b/sygnal/helper/context_factory.py
@@ -19,13 +19,13 @@
 
 import logging
 
-import idna
 from OpenSSL import SSL
 from service_identity import VerificationError
 from service_identity.pyopenssl import verify_hostname, verify_ip_address
 from twisted.internet.abstract import isIPAddress, isIPv6Address
 from twisted.internet.interfaces import IOpenSSLClientConnectionCreator
 from twisted.internet.ssl import CertificateOptions, TLSVersion, platformTrust
+from twisted.protocols.tls import TLSMemoryBIOProtocol
 from twisted.python.failure import Failure
 from twisted.web.iweb import IPolicyForHTTPS
 from zope.interface import implementer
@@ -43,7 +43,7 @@ class ClientTLSOptionsFactory:
     constructs an SSLClientConnectionCreator factory accordingly.
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
         # Use CA root certs provided by OpenSSL
         trust_root = platformTrust()
 
@@ -61,13 +61,13 @@ class ClientTLSOptionsFactory:
         self._verify_ssl_context = self._verify_ssl.getContext()
         self._verify_ssl_context.set_info_callback(self._context_info_cb)
 
-    def get_options(self, host):
+    def get_options(self, host: bytes) -> IOpenSSLClientConnectionCreator:
         ssl_context = self._verify_ssl_context
 
         return SSLClientConnectionCreator(host, ssl_context)
 
     @staticmethod
-    def _context_info_cb(ssl_connection, where, ret):
+    def _context_info_cb(ssl_connection: SSL.Connection, where: int, ret: int) -> None:
         """The 'information callback' for our openssl context object."""
         # we assume that the app_data on the connection object has been set to
         # a TLSMemoryBIOProtocol object. (This is done by SSLClientConnectionCreator)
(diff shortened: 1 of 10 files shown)


sygnal/helper/context_factory.py
```diff
@@ -125,22 +129,19 @@ class ConnectionVerifier:
 
     # This code is based on twisted.internet.ssl.ClientTLSOptions.
 
-    def __init__(self, hostname):
-        if isIPAddress(hostname) or isIPv6Address(hostname):
-            self._hostnameBytes = hostname.encode("ascii")
+    def __init__(self, hostname: bytes):
+        _decoded = hostname.decode("ascii")
+        if isIPAddress(_decoded) or isIPv6Address(_decoded):
```
DMRobertson, line 134: Confusingly these functions [will accept bytes](https://github.com/twisted/twisted/blob/525da4a0becad1e814c74f12c365f15fe9cd10de/src/twisted/internet/abstract.py#L487-L503), even though the annotations disallow that.

sygnal/helper/context_factory.py
```diff
@@ -125,22 +129,19 @@ class ConnectionVerifier:
...
             self._is_ip_address = True
         else:
-            # twisted's ClientTLSOptions falls back to the stdlib impl here if
-            # idna is not installed, but points out that lacks support for
-            # IDNA2008 (http://bugs.python.org/issue17305).
-            #
-            # We can rely on having idna.
-            self._hostnameBytes = idna.encode(hostname)
```
DMRobertson, line 138: I'm not sure this is safe to remove---this looks like it's effectively removing support for internationalised domain names. What is your thinking here?

sygnal/helper/proxy/connectproxyclient_twisted.py
```diff
@@ -187,7 +194,7 @@ def connectionLost(self, reason=connectionDone):
         if not self.connected_deferred.called:
             self.connected_deferred.errback(reason)
 
-    def proxyConnected(self, _):
+    def proxyConnected(self, _: Union[None, "defer.Deferred[None]"]) -> None:
```
DMRobertson, line 197:
```suggestion
    def proxyConnected(self, _: Optional["defer.Deferred[None]"]) -> None:
```

Might need to import Optional if it's not already imported in this file.

sygnal/helper/proxy/proxyagent_twisted.py
```diff
@@ -89,7 +95,13 @@ def __init__(
...
         self._reactor = reactor
 
-    def request(self, method, uri, headers=None, bodyProducer=None):
+    def request(
+        self,
+        method: bytes,
+        uri: bytes,
+        headers: Optional[Headers] = None,
+        bodyProducer: Optional[IBodyProducer] = None,
+    ) -> "defer.Deferred[IResponse]":
```
DMRobertson, line 104: Could you remove the types from the docstring? We generally prefer to have the types specified in the signature only, to prevent the types in docstring from going out of sync.

DMRobertson (changes requested):
Many thanks. Just a few comments:

sygnal/helper/proxy/proxyagent_twisted.py
dklimpel: Done

sygnal/helper/context_factory.py
dklimpel: I have returned it.

sygnal/helper/proxy/connectproxyclient_twisted.py
dklimpel: Done. I thought it had a deeper purpose.
https://github.com/matrix-org/synapse/blob/aec69d2481e9ea1d8ea1c0ffce1706a65a7896a8/synapse/http/connectproxyclient.py#L213

sygnal/helper/proxy/proxyagent_twisted.py
```diff
@@ -113,20 +112,20 @@ def request(
         See also: twisted.web.iweb.IAgent.request
 
         Args:
-            method (bytes): The request method to use, such as `GET`, `POST`, etc
+            method: The request method to use, such as `GET`, `POST`, etc
 
-            uri (bytes): The location of the resource to request.
+            urip: The location of the resource to request.
```
DMRobertson:
```suggestion
            uri: The location of the resource to request.
```

sygnal/helper/proxy/connectproxyclient_twisted.py
DMRobertson: No deeper purpose here: `Optional[T]` is literally a alias of `Union[T, None]`. (In more modern versions of Python, it's spelt `T | None`, but the `|` operator for types wasn't introduced until Python 3.9, I think.)

sygnal/helper/proxy/proxyagent_twisted.py
DMRobertson: Oh, I think this was fixed in a later commit.

DMRobertson (approved):
Many thanks!

Issue #301, opened 2022-04-05
APNS device token get 400 for hexadecimal (original base64 ok).


timabilov:
**Describe the bug**
For APNS app i am unable to get notification if  hexadecimal representation of the token used. 

If in this code i change
```device_token = base64.b64decode(device.pushkey).hex() ->  device_token = device.pushkey ```
then it **works** as expected.  I know that from  Troubleshoot it was because of the backward compatibility, but then what i am doing wrong here?  To be clear, by decoding and encoding my device token(base64.b64..)  i do get my initial token.
Can it be related to that?
https://developer.apple.com/news/?id=uzyxiriy

**To Reproduce**
1. Get any base64 token from device
2. Call sygnal api with this base64 token
3. you will get BadDeviceToken 

**Expected behavior**
Device should get the notification

Device IPhone 12 Pro Max

used p8 certificate sandbox environment

timabilov:
https://github.com/matrix-org/sygnal/pull/96/commits

this is the related issue,
how is that not fixed yet?

timabilov:
it is react native push notification ios library which already does hexadecimal convert

PR #344, opened 2023-08-17
Allow to configure if pushkey should be converted to hex before sending to APNs

danbim:
I ran into the same problem as https://github.com/matrix-org/sygnal/issues/301 and decided to instead make this configurable in Sygnal. Not sure if this is the right approach, happy for feedback and adopting to something else :)

Signed-off-by: Daniel Bimschas <daniel@bimschas.com> / <daniel.bimschas@inoio.de>

### DIFF
--- a/sygnal.yaml.sample
+++ b/sygnal.yaml.sample
@@ -197,6 +197,11 @@ apps:
   #  #
   #  # The default is 'production'. Uncomment to use the sandbox instance.
   #  #platform: sandbox
+  #  #
+  #  # Specifies whether to use a hexadecimally encoded push token. If True
+  #  # (default) the pushkey received is trancoded from base64 to hex. If False
+  #  # the pushkey received is passed to APNs as is.
+  #  #apns_device_token_use_hex: false
 
   # This is an example GCM/FCM push configuration.
   #
--- a/sygnal.yaml.sample
+++ b/sygnal.yaml.sample
@@ -198,9 +198,9 @@ apps:
   #  # The default is 'production'. Uncomment to use the sandbox instance.
   #  #platform: sandbox
   #  #
-  #  # Specifies whether to use a hexadecimally encoded push token. If True
-  #  # (default) the pushkey received is trancoded from base64 to hex. If False
-  #  # the pushkey received is passed to APNs as is.
+  #  # Specifies whether to convert the device push token from base 64 to hex.
+  #  # Defaults to True, set this to False if your client library provides a
+  #  # push token in hex format.
   #  #apns_device_token_use_hex: false
 
   # This is an example GCM/FCM push configuration.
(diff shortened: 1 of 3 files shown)


sygnal.yaml.sample
```diff
@@ -197,6 +197,11 @@ apps:
   #  #
   #  # The default is 'production'. Uncomment to use the sandbox instance.
   #  #platform: sandbox
+  #  #
+  #  # Specifies whether to use a hexadecimally encoded push token. If True
+  #  # (default) the pushkey received is trancoded from base64 to hex. If False
+  #  # the pushkey received is passed to APNs as is.
+  #  #apns_device_token_use_hex: false
```
reivilibre, line 204: this option seems like it is named the opposite of what it means: I would expect `use_hex = True` to mean that Sygnal expects hex tokens from clients rather than base64.

Is it just me?

reivilibre (changes requested):
Happy with the idea of the change but just the option seems to be called the opposite of what it should be, if it's me being stupid then do say :)

sygnal.yaml.sample
danbim: Hi @reivilibre how about naming it `apns_device_token_transcode_to_hex`?

clokep: I think the `apns` is probably redundant since it is in the apns pushkin config, maybe `convert_device_token_to_hex`?

I suspect the comment needs to be updated a bit too:

> Whether to convert the device push token from base 64 to hex. Defaults to True, set this to False if your client library provides a push token in hex format.

sygnal/apnspushkin.py
```diff
@@ -242,7 +242,10 @@ async def _dispatch_request(
         log.info(f"Sending as APNs-ID {notif_id}")
         span.set_tag("apns_id", notif_id)
 
-        device_token = base64.b64decode(device.pushkey).hex()
+        if self.get_config("apns_device_token_use_hex", bool, True):
```
clokep, line 245:
```suggestion
        # Some client libraries will provide the push token in hex format already. Avoid
        # attempting to convert from base 64 to hex.
        if self.get_config("apns_device_token_use_hex", bool, True):
```

Or something like that.

sygnal.yaml.sample
clokep: I don't have a strong opinion on the prefix, I'll note that none of the other APNS-specific options have it though.

clokep (dismissed):

clokep:
Would you mind adding a newsfile?

clokep (approved):

clokep:
Will need to ask @matrix-org/synapse-core id they have another release planned soon.

clokep:
@danbim You'll need to ask the current team -- maybe @erikjohnston or @reivilibre would know.

Half-Shot:
This appears to have now happened -> https://github.com/matrix-org/sygnal/releases/tag/v0.14.0

PR #343, opened 2023-07-15
Bump python version

dklimpel:
Python 3.7 is EOL

Python 3.11 is also the Python version of Synapse in Dockerfile

Max Python version here: 3.10.
The upgrade to Python 3.11 seems to need more dependency updates.

```console
      gcc -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -Iast27/Include -I/home/runner/work/sygnal/sygnal/.tox/check_codestyle/include -I/opt/hostedtoolcache/Python/3.11.4/x64/include/python3.11 -c ast27/Custom/typed_ast.c -o build/temp.linux-x86_64-cpython-311/ast27/Custom/typed_ast.o
      In file included from ast27/Custom/typed_ast.c:3:
      ast27/Custom/../Include/compile.h:5:10: fatal error: code.h: No such file or directory
          5 | #include "code.h"
            |          ^~~~~~~~
      compilation terminated.
      error: command '/usr/bin/gcc' failed with exit code 1
```
Possible solutions / causes:
- https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1878635.html
- https://github.com/pre-commit/pre-commit/issues/2850#issuecomment-1519727166

Related to:
- #342

PR #347, opened 2023-09-05
Update aioapns dependency to one that supports python 3.10+

bradtgmurray:
I believe apns support has been broken since https://github.com/matrix-org/sygnal/pull/343, see comment https://github.com/matrix-org/sygnal/pull/343#issuecomment-1707191931

This brings us to the latest version of aioapns and fixes it to work on newer version of Python.

### DIFF
--- a/changelog.d/347.misc
+++ b/changelog.d/347.misc
@@ -1,0 +1,1 @@
+Bump aioapns dependency to 3.0 in order to support Python 3.10+
--- a/changelog.d/347.misc
+++ b/changelog.d/347.misc
@@ -1,1 +1,1 @@
-Bump aioapns dependency to 3.0 in order to support Python 3.10+
+Bump aioapns dependency to 3.0 in order to support Python 3.10+.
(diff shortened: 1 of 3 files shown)


changelog.d/347.misc
clokep, line 1: Just to convince myself this was safe, it seems that we [really need >= 2.1](https://github.com/Fatal1ty/aioapns/releases/tag/v2.1) to support Python 3.10.

Bumping to 3.0 itself is probably fine though.

clokep (approved):
Looking quickly I don't see anything else useful in 3.0. 🤷

clokep:
@bradtgmurray Thanks! I just realized you didn't sign-off on these changes, would you mind adding your sign-off when you get a moment?

clokep:
> @bradtgmurray Thanks! I just realized you didn't sign-off on these changes, would you mind adding your sign-off when you get a moment?

The description was updated, so we're all set. 👍 Thanks again!

Issue #342, opened 2023-07-15
Upgrade dependencies


sephii:
Some dependencies are quite old:

* prometheus-client: 0.7.1 (2019-06-20) vs 0.17.1 (2023-07-10)
* matrix-common: 1.0.0 (2021-12-03) vs 1.3.0 (2022-09-13)
* aioapns: 2.0.2 (2021-02-03) vs 2.2 (2022-11-11)

It would be nice to update them to make packaging Sygnal easier.

reivilibre:
I reckon we'd be happy to accept contributions that do this, as long as the new versions are tested to work. At the moment I would, if being frank, say this is unlikely to happen soon without being contributed.

matrixbot:
For your information, this issue has been copied over to the Element fork of `sygnal`: <https://github.com/element-hq/sygnal/issues/342>

PR #348, opened 2023-09-30
Bump `matrix-common` to `1.3.0`

dklimpel:
Related to:
- #342


clokep (approved):
Should be fine, 1.2 was mostly packaging changes. 1.3 adds some support for MXC URLs which isn't helpful here.

PR #349, opened 2023-09-30
Bump `mypy`, `black` and `isort`

dklimpel:
Related to:

- #342

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,12 +94,12 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    "black==22.3.0",
+    "black>=22.7.0",
     "coverage~=5.5",
     "ruff==0.0.262",
     "isort~=5.0",
-    "mypy==0.812",
-    "mypy-zope==0.3.0",
+    "mypy",
+    "mypy-zope",
     "towncrier",
     "tox",
     "types-opentracing>=2.4.2",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -94,15 +94,17 @@ dependencies = [
...
-    "black==22.3.0",
+    "black>=22.7.0",
     "coverage~=5.5",
     "ruff==0.0.262",
-    "isort~=5.0",
-    "mypy==0.812",
-    "mypy-zope==0.3.0",
+    "isort>=5.10.1",
+    "mypy",
+    "mypy-zope",
```
clokep, line 102: Note that the reason we don't pin mypy, mypy-zope, and black in the Synapse config here is that we assume developers are using the `poetry.lock` file.

Since we don't use that for sygnal (yet?) we should probably just pin versions here.

clokep: (I think that makes sense at least, @DMRobertson any thoughts?)

DMRobertson: Yeah, we want to pin the CI/linter tooling _somewhere_ so that CI doesn't break underneath us.

pyproject.toml
```diff
@@ -94,15 +94,17 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    "black==22.3.0",
+    "black==23.9.1",
     "coverage~=5.5",
     "ruff==0.0.262",
```
clokep, line 99: Should we bump `ruff` too while we're at it?

clokep (approved):
Seems good, let me know if you want to do ruff also.

clokep (approved):
Thank you!

PR #358, opened 2024-01-26
Add a new APNs configuration option `push_with_badge` to suppress sending `aps.badge` in APNs messages.

CEnnis91:
For situations when not using `mutable-content` and updating the badge count is not desired.
This also adds `unread_count` and `missed_calls` to the push payload when `push_with_badge` is `False`.

### DIFF
--- a/sygnal.yaml.sample
+++ b/sygnal.yaml.sample
@@ -202,6 +202,10 @@ apps:
   #  # Defaults to True, set this to False if your client library provides a
   #  # push token in hex format.
   #  #convert_device_token_to_hex: false
+  #  #
+  #  # Specifies whether to send the badge key in the APNs message.
+  #  # Defaults to True, set this to False to suppress sending the badge count.
+  #  #push_with_badge: false
 
   # This is an example GCM/FCM push configuration.
   #
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -110,6 +110,7 @@ class ApnsPushkin(ConcurrencyLimitedPushkin):
         "keyfile",
         "topic",
         "push_type",
+        "push_with_badge",
     } | ConcurrencyLimitedPushkin.UNDERSTOOD_CONFIG_FIELDS
 
     APNS_PUSH_TYPES = {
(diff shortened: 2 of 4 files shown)


sygnal/apnspushkin.py
```diff
@@ -551,14 +552,20 @@ def _get_payload_full(
         if loc_args:
             payload["aps"].setdefault("alert", {})["loc-args"] = loc_args
 
-        if badge is not None:
+        if self.get_config("push_with_badge", bool, True) and badge is not None:
```
anoadragon453, line 555: Rather than pulling this option value out of the config on each request, please do so in `__init__` and set to a class variable, then reference that class variable instead. i.e.:

```py
def __init__(self, name: str, sygnal: "Sygnal", config: Dict[str, Any]) -> None:
    # ...
    self.push_with_badge = self.get_config("push_with_badge", bool, True)
```

```py
def _get_payload_event_id_only(
    self, n: Notification, default_payload: Dict[str, Any]
) -> Dict[str, Any]:
    # ...
    if self.push_with_badge and badge is not None:
```

anoadragon453 (changes requested):

sygnal.yaml.sample
```diff
@@ -202,6 +202,10 @@ apps:
   #  # Defaults to True, set this to False if your client library provides a
   #  # push token in hex format.
   #  #convert_device_token_to_hex: false
+  #  #
+  #  # Specifies whether to send the badge key in the APNs message.
+  #  # Defaults to True, set this to False to suppress sending the badge count.
+  #  #push_with_badge: false
```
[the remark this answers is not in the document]
CEnnis91: I agree this comes off as a bit confusing.

The primary focus of this commit to suppress the `aps.badge` field so iOS does not automatically draw the badge on the home screen. In the event an iOS app is receiving APNS from multiple independent sources, the badge value is incorrect. Whether the other source is sending its own `aps.badge` or not, sygnal sending its own `aps.badge` still conflicts with the true badge value expected by the user.

When suppressing that field, I wanted to ensure the underlying data was still available to the iOS client, so it could be used, if desired. The inclusion of the `unread_count` and `missed_calls` fields were to represent the expanded form of what `aps.badge` represents [here](https://github.com/matrix-org/sygnal/blob/c8a85fd3e9009ff21a79089a4f3524cc0e310024/sygnal/apnspushkin.py#L532-L538) .

I agree the inclusion of `unread_count` and `missed_calls` could become their own setting, but I was unsure about APNS message length assumptions.

- If length assumptions are not a concern, it seems like always including `unread_count` and `missed_calls` in APNS would be reasonable (I can submit a separate PR). Then this PR would simply be about toggling the inclusion of `aps.badge`.
- If length assumptions are a concern, I'm open to opinions of how to move forward. We can drop the extra fields entirely, or add multiple settings controlling the inclusion of fields. But neither of these options feel very elegant.

sygnal/apnspushkin.py
CEnnis91: :+1: will be in the next commit

PR #360, opened 2024-02-08
Explicitly configure the aioapns event loop so proxying works when configured

devonh:
When the aioapns dependency was updated, they removed the ability to inject an event loop.
This change sets the event loop manually after creating the apns client.

We use a custom event loop in this case in order to override `create_connection` when a proxy is configured to force all traffic over the proxy.

Fixes https://github.com/element-hq/backend-internal/issues/22

### DIFF
--- a/sygnal/apnspushkin.py
+++ b/sygnal/apnspushkin.py
@@ -213,6 +213,9 @@ class ApnsPushkin(ConcurrencyLimitedPushkin):
         # without this, aioapns will retry every second forever.
         self.apns_client.pool.max_connection_attempts = 3
 
+        # without this, aioapns will not use the proxy if one is configured.
+        self.apns_client.pool.loop = loop
+
     def _report_certificate_expiration(self, certfile: str) -> None:
         """Export the epoch time that the certificate expires as a metric."""
         with open(certfile, "rb") as f:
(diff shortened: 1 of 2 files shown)


sygnal/apnspushkin.py
```diff
@@ -213,6 +213,9 @@ async def make_apns() -> aioapns.APNs:
         # without this, aioapns will retry every second forever.
         self.apns_client.pool.max_connection_attempts = 3
 
+        # without this, aioapns will not use the proxy if one is configured.
+        self.apns_client.pool.loop = loop
```
erikjohnston, line 217: I'm slightly hesitant to gut wrench into the depths of classes TBH, though appreciate this might be the best option.

Given this proxy seems to be a global value, I wonder if we could create the new loop at startup and do `asyncio.set_event_loop(loop)`?

devonh: I tried doing that, in many many different ways but couldn't get aioapns to actually behave and use that loop.
So I gave up and resorted to this method (which I also don't like)

Maybe there is a way to do it with setting the event loop, but I wasn't able to make it work.

erikjohnston: Huh, wtf. Having looked at the code it seems that it just does `asyncio.get_event_loop()` in the constructor, but maybe something is somehow constructing a class at import time or something? Weird.

erikjohnston (approved):

PR #361, opened 2024-02-12
Support for FCM v1 API

devonh:
This adds support for the new FCM v1 API.
Major changes introduced:
- can only send to a single device per request, no more batch sending
- push data format changes, only `string: string` entries are allowed in the data section
- small config tweaks


erikjohnston (approved):
LGTM! Thanks! Have we tested this in the wild at all?

octopop:
@devonh What is the ETA or plan to release these changes as a tagged release and updated docker image?

PR #362, opened 2024-03-08
Check if `data["content"]` is a dict before using it as a dict

c-cal:
Fix https://github.com/matrix-org/sygnal/issues/363

image

When notification does not contain the "content" field, then `None` is used by default:
https://github.com/matrix-org/sygnal/blob/c8a85fd3e9009ff21a79089a4f3524cc0e310024/sygnal/notifications.py#L95

This happens with pusher that have the attribute `data.format: event_id_only`, such as Element Android (https://github.com/element-hq/element-android/blob/develop/vector/src/main/java/im/vector/app/core/pushers/PushersManager.kt#L85), or when Synapse is configured with `push.include_content: false`.

Signed-off-by: Charlie Calendre

### DIFF
--- a/changelog.d/362.bugfix
+++ b/changelog.d/362.bugfix
@@ -1,0 +1,1 @@
+Fix a bug causing Sygnal to fail when processing notifications without a `content` dict, when those notifications were destined for GCM. Contributed by @c-cal.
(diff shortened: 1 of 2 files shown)


reivilibre (approved):
Thanks!

PR #373, opened 2024-05-15
Update docker build and workflows to Python 3.11

MatMaul:
Python 3.11 is now old enough to be considered pretty stable I believe, and it has major perfs improvement.

Moreover it supports [TLS-in-TLS](https://github.com/aio-libs/aiohttp/discussions/6044), enabling support for HTTPS proxy for FCM v1 API, which use `google-auth` with `aiohttp`.


devonh (approved):
Thanks for this!

PR #374, opened 2024-05-16
Switch to using poetry

devonh:
This brings sygnal in line with synapse & allows the use of a poetry.lock file to track dependency versions more precisely.


erikjohnston (changes requested):
Thank you! This looks good, but I think we need to update the dockerfile to use poetry?

devonh:
> Thank you! This looks good, but I think we need to update the dockerfile to use poetry?

Ah! I knew I must be missing something. Will do.

docker/Dockerfile
```diff
@@ -6,26 +6,81 @@
...
+###
+### Stage 0: generate requirements.txt
+###
+# We hardcode the use of Debian bookworm here because this could change upstream.
+FROM docker.io/library/python:${PYTHON_VERSION}-slim-bookworm as requirements
+
+# We install poetry in its own build stage to avoid its dependencies conflicting with
+# sygnal's dependencies.
+RUN --mount=type=cache,target=/root/.cache/pip \
+  pip install --user "poetry==1.3.2"
```
erikjohnston, line 20: That's quite an old version of poetry to pin to? It's now at 1.8.3

erikjohnston (commented):
Otherwise looks good. I assume you've manually tested building and running the docker image?

docker/Dockerfile
devonh: I just grabbed the same version from synapse. I'll bump it up here since I don't think we have any of the same constraints.

PR #383, opened 2024-06-10
Use google `default` for auth when no `service_account_json` file specified

devonh:
This allows the use of various other mechanisms to acquire google application credentials.
See here for more info:
https://cloud.google.com/docs/authentication#auth-decision-tree
https://cloud.google.com/docs/authentication/application-default-credentials

To achieve the same behaviour as specifying `service_account_json` in the config, set the env var `GOOGLE_APPLICATION_CREDENTIALS="/path/to/service_account.json"` 

Note: This has the downside of applying the same google credentials for all GCM pushers.
If you desire different google credentials for different pushkins running on the same Sygnal instance, setting the `service_account_json` is the only way to achieve that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ API v1 also expects one of the following to be configured:
 
 - the `service_account_file` config parameter to contain the path to the service account file
 
-It is recommended to use the application default credentials method to acquire google credentials.
+It is recommended to use the application default credentials method to acquire Google credentials.
 
 Using an HTTP Proxy for outbound traffic
 ----------------------------------------
--- a/README.md
+++ b/README.md
@@ -99,9 +99,17 @@ For API v1, it expects:
 - the `project_id` parameter to contain the `Project ID`,
   which can be acquired from Firebase Console at:
   `https://console.cloud.google.com/project/<PROJECT NAME>/settings/general/`
+
+API v1 also expects one of the following to be configured:
 - the [Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials) configured on the local system.
   This can point to a service account file which can be acquired from Firebase Console at:
   `https://console.firebase.google.com/project/<PROJECT NAME>/settings/serviceaccounts/adminsdk`
+
+  **OR..**
+
+- the `service_account_file` config parameter to contain the path to the service account file
+
+It is recommended to use the application default credentials method to acquire google credentials.
 
 Using an HTTP Proxy for outbound traffic
 ----------------------------------------
--- a/README.md
+++ b/README.md
@@ -99,8 +99,8 @@ For API v1, it expects:
 - the `project_id` parameter to contain the `Project ID`,
   which can be acquired from Firebase Console at:
   `https://console.cloud.google.com/project/<PROJECT NAME>/settings/general/`
-- the `service_account_file` parameter to contain the path to the service account file,
-  which can be acquired from Firebase Console at:
+- the [Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials) configured on the local system.
+  This can point to a service account file which can be acquired from Firebase Console at:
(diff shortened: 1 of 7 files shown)


sygnal/gcmpushkin.py
```diff
@@ -186,23 +186,39 @@ def __init__(self, name: str, sygnal: "Sygnal", config: Dict[str, Any]) -> None:
...
+        if not isinstance(self.base_request_body, dict):
+            raise PushkinSetupException(
+                "Config field fcm_options, if set, must be a dictionary of options"
+            )
+
+    def _load_credentials(self, proxy_url: Optional[str]) -> None:
         self.credentials: Optional[Credentials] = None
 
         if self.api_version is APIVersion.V1:
             self.service_account_file = self.get_config("service_account_file", str)
```
anoadragon453, line 204: We seem to be trying the `service_account_file` config option first, before attempting the Application Default Credentials route. This seems counter-intuitive to me, given that environment variables typically trump config options when defined. And we recommend Application Default Credentials over the `service_account_file` config option.

tests/test_gcm.py
```diff
@@ -120,6 +117,12 @@ def preload_with_response(
         self.preloaded_response = DummyResponse(code)
         self.preloaded_response_payload = response_payload
 
+    def _load_credentials(self, proxy_url: Optional[str]) -> None:
+        if self.api_version is APIVersion.V1:
+            self.credentials = TestCredentials()  # type: ignore[assignment]
+            self.google_auth_request = None  # type: ignore[assignment]
+            pass
```
anoadragon453, line 124: Why is `pass` (a no-op) needed here?

tests/test_gcm.py
```diff
@@ -165,14 +151,10 @@ def config_setup(self, config: Dict[str, Any]) -> None:
...
             "fcm_options": {"content_available": True, "mutable_content": True},
         }
-        self.service_account_file = tempfile.NamedTemporaryFile()
-        self.service_account_file.write(FAKE_SERVICE_ACCOUNT_FILE)
-        self.service_account_file.flush()
         config["apps"]["com.example.gcm.apiv1"] = {
             "type": "tests.test_gcm.TestGcmPushkin",
             "api_version": "v1",
             "project_id": "example_project",
-            "service_account_file": self.service_account_file.name,
```
anoadragon453, line 175: Given we still support the `service_account_file` config option, it feels odd to remove it from the tests completely.

Perhaps we should leave it as a possible variant of this class? And have a simple test which check that variant loads credentials correctly?

README.md
```diff
@@ -99,10 +99,18 @@ For API v1, it expects:
...
+API v1 also expects one of the following to be configured:
+- the [Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials) configured on the local system.
+  This can point to a service account file which can be acquired from Firebase Console at:
   `https://console.firebase.google.com/project/<PROJECT NAME>/settings/serviceaccounts/adminsdk`
 
+  **OR..**
+
+- the `service_account_file` config parameter to contain the path to the service account file
+
+It is recommended to use the application default credentials method to acquire google credentials.
```
anoadragon453, line 112:
```suggestion
It is recommended to use the application default credentials method to acquire Google credentials.
```

anoadragon453 (commented):
A few questions, but nothing major.

sygnal/gcmpushkin.py
devonh: We could try the Application Default Credentials route first, but that leads to it trying for ~5 seconds before timing out. I didn't want to make startup longer if you had explicitly configured the `service_account_file`

devonh:
After thinking about this more, we need to be able to specify the service account file per pusher. Which means a global environment variable does not fit the bill.
Closing this PR.

PR #385, opened 2024-06-20
Change package name back to matrix-sygnal

devonh:
The package name was accidentally changed in #374 
This PR changes it back.


devonh:
Since we are now using poetry, poetry requires the top level package folder to match the package name.
So in order to rename the package to `matrix-sygnal`, I had to move the `sygnal` directory to `matrix_sygnal`. This results in a change to how sygnal is run. 

With this change, you need to run sygnal using:
`python -m matrix_sygnal.sygnal`

This could be even more of a footgun than having a new package name. (Having a new package name of `sygnal` means existing installs could end up with multiple copies of Sygnal in a virtualenv since the old package name was `matrix-sygnal` and doesn't conflict with the current name)

devonh:
> Since we are now using poetry, poetry requires the top level package folder to match the package name. So in order to rename the package to `matrix-sygnal`, I had to move the `sygnal` directory to `matrix_sygnal`. This results in a change to how sygnal is run.
> 
> With this change, you need to run sygnal using: `python -m matrix_sygnal.sygnal`
> 
> This could be even more of a footgun than having a new package name. (Having a new package name of `sygnal` means existing installs could end up with multiple copies of Sygnal in a virtualenv since the old package name was `matrix-sygnal` and doesn't conflict with the current name)

This has been resolved & is no longer true.
You can now continue to run using `python -m sygnal.sygnal`

erikjohnston (dismissed):

PR #386, opened 2024-06-20
Truncate large values inside of Firebase notification body fields.

devonh:
This PR changes the use of the Firebase v1 API to truncate the `content` fields within `data` to a maximum length.

NOTE: The existing code for the legacy API does not effectively truncate `content` since it relies on `len(data[attr]) > MAX_BYTES_PER_FIELD` which will return the number of keys inside the `content` dict. Since the legacy API is being removed tomorrow, I don't want to spend time coming up with a solution for this. At first glance it appears to be more involved than I would like to accurately count the size in bytes of the contents of the values inside of a python dict.

### DIFF
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -745,12 +745,26 @@ class GcmPushkin(ConcurrencyLimitedPushkin):
         return data
 
 
-def truncate_str(input: str, max_length: int) -> tuple[str, bool]:
+def truncate_str(input: str, max_bytes: int) -> tuple[str, bool]:
+    """
+    Truncate the given string. If the truncation would occur in the middle of a unicode
+    character, that character will be removed entirely instead.
+    Appends a `…` character to imply the string was truncated.
+
+    Args:
+        `input`: the string to be truncated
+        `max_bytes`: maximum length, in bytes, that the payload should occupy when truncated
+
+    Returns:
+        Tuple of (truncated string, whether truncation took place)
+
+    """
+
     str_bytes = input.encode("utf-8")
-    if len(str_bytes) <= max_length:
+    if len(str_bytes) <= max_bytes:
         return (input, False)
 
     try:
-        return (str_bytes[: max_length - 3].decode("utf-8") + "…", True)
+        return (str_bytes[: max_bytes - 3].decode("utf-8") + "…", True)
     except UnicodeDecodeError as err:
         return (str_bytes[: err.start].decode("utf-8") + "…", True)
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -705,18 +705,21 @@ class GcmPushkin(ConcurrencyLimitedPushkin):
                 data[attr] = getattr(n, attr)
                 # Truncate fields to a sensible maximum length. If the whole
                 # body is too long, GCM will reject it.
-                if data[attr] is not None and len(data[attr]) > MAX_BYTES_PER_FIELD:
-                    overflow_fields += 1
-                    data[attr] = data[attr][0:MAX_BYTES_PER_FIELD]
+                if data[attr] is not None and isinstance(data[attr], str):
+                    # The only `attr` that shouldn't be of type `str` is `content`,
+                    # which is handled explicitly later on.
+                    if len(bytes(data[attr], "utf-8")) > MAX_BYTES_PER_FIELD:
+                        overflow_fields += 1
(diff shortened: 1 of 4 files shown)


sygnal/gcmpushkin.py
```diff
@@ -707,6 +707,8 @@ def _build_data(
                 for attr, value in data["content"].items():
                     if not isinstance(value, str):
                         continue
+                    if len(value) > MAX_BYTES_PER_FIELD:
+                        value = value[0:MAX_BYTES_PER_FIELD]
```
erikjohnston, line 711: This will do for now, but we should a) handle this in unicode, and b) and an ellipsis at the end.

erikjohnston (dismissed):

sygnal/gcmpushkin.py
devonh: Ah - I had just borrowed this from the other places we already do truncation. 
I'll update this to add an ellipsis, but leave the unicode for a future PR.

sygnal/gcmpushkin.py
```diff
@@ -700,13 +706,17 @@ def _build_data(
...
                     data[attr] = data[attr][0:MAX_BYTES_PER_FIELD]
 
         if api_version is APIVersion.V1:
             if isinstance(data.get("content"), dict):
                 for attr, value in data["content"].items():
                     if not isinstance(value, str):
                         continue
+                    if len(value) > MAX_BYTES_PER_FIELD:
+                        overflow_fields += 1
+                        value = value[0:MAX_BYTES_PER_FIELD] + "..."
```
erikjohnston, line 719:
```suggestion
                        value = value[:MAX_BYTES_PER_FIELD-3] + "…"
```

(The ellipsis is 3 bytes either way).

sygnal/gcmpushkin.py
```diff
@@ -700,13 +706,17 @@ def _build_data(
...
                 if data[attr] is not None and len(data[attr]) > MAX_BYTES_PER_FIELD:
+                    overflow_fields += 1
                     data[attr] = data[attr][0:MAX_BYTES_PER_FIELD]
 
         if api_version is APIVersion.V1:
             if isinstance(data.get("content"), dict):
                 for attr, value in data["content"].items():
                     if not isinstance(value, str):
                         continue
+                    if len(value) > MAX_BYTES_PER_FIELD:
```
erikjohnston, line 717: Are these bytes or strings? Looking at it python will count code points for strings.

tests/test_gcm.py
```diff
@@ -480,3 +480,105 @@ def test_fcm_options(self) -> None:
...
+xxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxx\
+ooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxoooooooooo\
+xxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxx\
+ooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxoooooooooo\
+xxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxx\
+ooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxoooooooooo\
+xxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxx\
+ooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxoooooooooo\
+xxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxxxxxxxx\
+ooooooooooxxxxxxxxxxooooooooooxxxxxxxxxxooooooooooxxxx...",
```
erikjohnston, line 551: Might be good to add a unicode test case

sygnal/gcmpushkin.py
devonh: I had the subtraction initially but then took it out. The MAX_BYTES is pretty arbitrary as it is.
I'm happy either way.

sygnal/gcmpushkin.py
devonh: Ah, they are indeed code points. Good catch.

erikjohnston (approved):

PR #394, opened 2024-10-01
Only allow aiohttp < 3.10.0

V02460:
Newer versions of aiohttp break Sygnal when using FCM + proxy. Since aiohttp>=3.10, the following error occurs: `RuntimeError: no running event loop`. IIUC, this is because Sygnal is [using aiohttp incorrectly](https://github.com/matrix-org/sygnal/blob/345aa61a59cebc2b87a7375ab696838f98e37097/sygnal/gcmpushkin.py#L217).

Downstream bug (German): https://gitlab.opencode.de/bwi/bundesmessenger/backend/helm-chart/-/issues/40

### DIFF
--- a/changelog.d/394.bugfix
+++ b/changelog.d/394.bugfix
@@ -1,0 +1,1 @@
+Only allow aiohttp < 3.10.0.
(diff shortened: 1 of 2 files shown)


anoadragon453:
Hi @V02460. Thanks for raising this and for the PR. I've gone ahead and fixed Sygnal's use of `aiohttp.ClientSession` in https://github.com/matrix-org/sygnal/pull/395, so the latest versions of `aiohttp` can be used.

PR #395, opened 2024-10-02
Fix breakage with `aiohttp>=3.10.0` when using GCM and an HTTP proxy

anoadragon453:
Since `aiohttp==3.10.0`, instantiating `aiohttp.ClientSession` must be done from an async function. Calling this from a non-async function relied on deprecated behaviour in Python, which broke in recent releases. https://github.com/aio-libs/aiohttp/issues/8555#issuecomment-2263814747 explains the situation.

We would do this when using GCM + an HTTP proxy:

https://github.com/matrix-org/sygnal/blob/345aa61a59cebc2b87a7375ab696838f98e37097/sygnal/gcmpushkin.py#L213-L217

Rather than pin `aiohttp` to an older version, let's just fix the behaviour. Supersedes https://github.com/matrix-org/sygnal/pull/394.

### DIFF
--- a/sygnal/gcmpushkin.py
+++ b/sygnal/gcmpushkin.py
@@ -239,6 +239,9 @@ class GcmPushkin(ConcurrencyLimitedPushkin):
             # `ClientSession` can't directly take the proxy URL, so we need to
             # set the usual env var and use `trust_env=True`
             os.environ["HTTPS_PROXY"] = proxy_url
+
+            # ClientSession must be instantiated by an async function, hence we do this
+            # here instead of `__init__`.
             session = aiohttp.ClientSession(trust_env=True, auto_decompress=False)
 
         cls.google_auth_request = google.auth.transport._aiohttp_requests.Request(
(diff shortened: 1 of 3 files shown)


sygnal/gcmpushkin.py
```diff
@@ -209,16 +209,10 @@ def __init__(self, name: str, sygnal: "Sygnal", config: Dict[str, Any]) -> None:
...
-                os.environ["HTTPS_PROXY"] = proxy_url
-                session = aiohttp.ClientSession(trust_env=True, auto_decompress=False)
-
-            self.google_auth_request = google.auth.transport._aiohttp_requests.Request(
-                session=session
-            )
+        # This is built in `_refresh_credentials`, and must be done in an async function.
+        self.google_auth_request: Optional[
+            google.auth.transport._aiohttp_requests.Request
+        ] = None
```
erikjohnston, line 215: Given this constructor is only called from the async function `GcmPushkin.create`, can we instead instansiate the google auth request there?

Otherwise it's not immediately clear to me what would happen if building the request failed?

anoadragon453: Oops, I had totally missed that. That simplifies things greatly, thanks! Used in https://github.com/matrix-org/sygnal/pull/395/commits/0cab5e6b48150105707c4e3d6d5c08c1560eea85.

erikjohnston: I only spotted it because I was going to suggest adding it :laughing:

erikjohnston (approved):

PR #409, opened 2025-04-04
Speed up the building of Docker images

AndrewFerr:
Analogous to element-hq/synapse#18038, switch from `pip install` to `uv pip install`, and use the build target for arch-agnostic stages. Also define build arguments for otherwise hardcoded values.


anoadragon453 (approved):
Changes look good and docker image builds locally. Thanks!

PR #410, opened 2025-04-23
Add basic instructions for running within a container.

pixlwave:
Nothing huge here, just adding a small example as I found myself doing this last week and didn't find any existing instructions for how to run Sygnal in a container.

Not sure if the changelog is really necessary here but added one in case :)


devonh (approved):
Thanks for adding this! I'm sure others will find it helpful in the future.

PR #413, opened 2025-05-12
Modify proxy test to use mocked FCM API and test it in CI

reivilibre:
This takes the proxy test introduced in #375 and:

- replaces the nginx forward proxy (which doesn't currently have an AMD64 `latest` release :S) with a [mitmproxy](https://mitmproxy.org) with some hardcoded dummy responses, so we don't need to configure FCM credentials and hit the real FCM API
    - so now you can run the proxy test with minimal faff
- makes `curl.sh` exit with a failure status if the Sygnal request doesn't give 200 OK
- adds CI code that spawns the Compose setup and then exercises `curl.sh`, resulting in a success or fail
    - this involves building the dockerfile in more situations than we did before, so caching is now enabled for the docker build
    - we also have to export and import the container image as an 'artifact' to ferry it across jobs


scripts-dev/proxy-test/docker-compose.yml
```diff
@@ -11,18 +13,28 @@ services:
...
       no-internet:
         ipv4_address: 172.28.0.3
       internet:
-    container_name: nginx-forward-proxy
+    container_name: mitmdump
     volumes:
+      - ./out:/out:Z
+      - ./mitmproxy:/home/mitmproxy/.mitmproxy:z
+      - ./responses:/responses:ro
       - ./nginx.conf:/usr/local/nginx/conf/nginx.conf:ro
```
devonh, line 38: I assume this and the `/scripts-dev/proxy-test/nginx.conf` file can also be removed?
```suggestion
```

.github/workflows/docker_check.yml
```diff
@@ -39,10 +67,86 @@ jobs:
...
+      - name: Invoke test script
+        run: |
+          echo '====='
+          docker ps -a
+          # TODO do something less hacky than a static 5-second sleep
+          sleep 5
+          echo '====='
+          docker ps -a
+          echo '====='
+          docker exec sygnal sh /curl.sh notification-gcm.json
```
devonh, line 135: Is this just waiting for sygnal to be reachable?
Should we be adding a `/health` endpoint to Sygnal and be relying on docker to wait for that?

devonh (commented):
Thanks for doing this, the changes look great!
Just a couple small comments.

.github/workflows/docker_check.yml
reivilibre: Good point. Have opened https://github.com/matrix-org/sygnal/pull/418 to add this.

reivilibre: although not entirely sure if the compose action waits for it to be healthy :SS

reivilibre: I added a loop that polls the healthiness (using healthcheck from https://github.com/matrix-org/sygnal/pull/418).

devonh (approved):
Awesome! Hopefully this squashes any future proxy issues from appearing in Sygnal without getting caught in CI.

PR #418, opened 2025-06-30
Add HEALTHCHECK directive to the Dockerfile

reivilibre:
Tested against `podman` (running `podman compose` flow from https://github.com/matrix-org/sygnal/pull/413). Verified that it reports as 'healthy'.

The endpoint was introduced in https://github.com/matrix-org/sygnal/pull/55 (Sygnal 0.2.3).

### DIFF
--- a/changelog.d/418.docker
+++ b/changelog.d/418.docker
@@ -1,1 +1,1 @@
-Add HEALTHCHECK directive to the container image, using existing `/health` endpoint.
+Add `HEALTHCHECK` directive to the container image, using existing `/health` endpoint.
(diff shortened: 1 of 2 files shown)


changelog.d/418.docker
```diff
@@ -0,0 +1 @@
+Add HEALTHCHECK directive to the container image, using existing `/health` endpoint.
```
MadLittleMods, line 1:
```suggestion
Add `HEALTHCHECK` directive to the container image, using existing `/health` endpoint.
```

MadLittleMods (approved):

